PR #2, opened 2025-01-17
Feature/automation

PhoenixCodingStuff:
Added automatic releases by using `[Release]` as prefix
Added automatic discord announcements use `[Skip]` as prefix to skip
Added automatic live tests [can expand later by making an addon for testing and automatically uploading and running commands] use `[Skip]` as prefix to skip
Added automatic versioning where build number is automatically changed
Made gradle build only if it isn't prefixed with `[Skip]` 
Limited stuff in order to skip unnecessary action runs [example run build and stuff only if its a valid change and not something like Update Readme]


build.gradle.kts
```diff
@@ -10,7 +10,9 @@ plugins {
     id("net.thebugmc.gradle.sonatype-central-portal-publisher") version "1.2.4"
 }
 
-group = "io.github.pylonmc"
+group = "io.github.pylon-paper"
```
LordIdra, line 13: pylon-paper?

.github/workflows/release.yml
```diff
@@ -0,0 +1,37 @@
+name: Create new Release
```
LordIdra, line 1: Are we going to link this with sonatype releases as well?

.github/workflows/live-testing.yml
```diff
@@ -0,0 +1,67 @@
+name: Live Testing
```
LordIdra, line 1: Nice this looks good. We'll have to figure out how to communicate success/failure from the tests tho

.github/workflows/release.yml
Seggan: Sounds good to me

.github/workflows/live-testing.yml
LordIdra: Yeah let's cross that bridge later. I'll need to make the test addon when blockstorage is done so that we can find out that blockstorage doesn't work. we can integrate stuff then.

build.gradle.kts
```diff
@@ -11,6 +11,8 @@ plugins {
 }
 
 group = "io.github.pylonmc"
+version = "UNKNOWN-SNAPSHOT"
```
Seggan, line 14: `version` is already specified in `gradle.properties`, if you wanna override the default go there

Seggan (changes requested):

JustAHuman-xD:
By the way, we should look into caching our live-testing setup, currently it'll be downloading the server jar and setting everything up *every* time which will be pretty slow, having it runnable on demand would be way better, I don't know how to cache it personally but I've seen it done in some project (gonna try and find it). This isn't something to prevent this PR being merged, just something to look into for the future

.github/workflows/gradle.yml
```diff
@@ -21,6 +27,10 @@ jobs:
         java-version: '21'
         distribution: 'temurin'
 
+    - id: env-setup
+      run: |
+        sed -i "s/version=\"-SNAPSHOT\"/version=\"${{ github.run_number }}-SNAPSHOT\"/g" pylon-core/gradle.properties
```
Seggan, line 32: Does this replace the version in gradle.properties? It's a lot better to simply pass it as a command arg with -Pversion=

.github/workflows/live-testing.yml
```diff
@@ -0,0 +1,67 @@
...
+
+      - name: Set up JDK 21
+        uses: actions/setup-java@v4.2.1
+        with:
+          distribution: 'adopt'
+          java-version: '21'
+          java-package: jdk
+          architecture: x64
+
+      - name: Setup server
```
Seggan, line 35: As I already said this needs to use the built in Gradle `runServer` task

Seggan (changes requested):
If needed I can do these myself on Monday

.github/workflows/discord-webhook.yml
```diff
@@ -0,0 +1,39 @@
...
+        uses: tsickert/discord-webhook@v6.0.0
+        with:
+          webhook-url: ${{ secrets.DISCORD_TOKEN }}
+          username: Pylon
+          embed-title: Pylon Update! - ${{ github.event.repository.name }}
+          embed-description: |
+            ${{ steps.description.outputs.id }}
+          embed-color: 5301186
+          embed-timestamp: ${{ github.event.head_commit.timestamp }}
+          embed-footer-text: Enjoy the update!
```
OhmV-IR, line 39: FYI, github has a built-in webhook feature that can send messages to discord. I think it's probably better to use that instead of this custom code. Example here: https://gist.github.com/jagrosh/5b1761213e33fc5b54ec7f6379034a22
this is the method I use for my projects

.github/workflows/gradle.yml
```diff
@@ -29,8 +39,14 @@ jobs:
     - name: Build with Gradle Wrapper
       run: ./gradlew build
 
-  dependency-submission:
+    - name: Upload the artifact
+      uses: actions/upload-artifact@v4.3.1
+      with:
+        name: pylon
+        path: 'build/libs/pylon-core-${{ github.run_number }}.jar'
```
OhmV-IR, line 46: Leaving this comment here to initiate discussion about how long we want to keep artifacts. There is a 2GB storage limit but it shouldn't be a big problem since each jar is only around 6MB. I would say 90 days(default) is probably an appropriate timeframe

.github/workflows/maven-central.yml
```diff
@@ -5,8 +5,46 @@ on:
...
+      uses: actions/upload-artifact@v4.3.1
+      with:
+        name: pylon
+        path: 'build/libs/pylon-core-${{ github.run_number }}.jar'
+
+  test-build:
+    needs: [build]
+    uses: ./.github/workflows/live-testing.yml
+    with:
+      artifact-name: pylon-core-${{ github.run_number }}
```
OhmV-IR, line 44: The build job is already done in another yml file, please merge this job and the one in gradle.yml into one file and call it using in L68-72 of gradle.yml(or maybe you can call the one that's in gradle.yml from here, idk).

.github/workflows/release.yml
```diff
@@ -0,0 +1,37 @@
...
+        required: true
+
+jobs:
+  release:
+    name: Create new release
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4.1.1
```
OhmV-IR, line 21: I don't believe this is necessary for this action

.github/workflows/release.yml
```diff
@@ -0,0 +1,37 @@
...
+    - name: Create Release
+      uses: marvinpinto/action-automatic-releases@latest
+      with:
+        repo_token: ${{ secrets.GITHUB_TOKEN }}
+        draft: false
+        prerelease: false
+        automatic_release_tag: v${{ github.run_number }}
+        title: Release ${{ github.run_number }}
+        files: |
+          *.jar
```
OhmV-IR, line 37: If possible, please specify the file path, don't want to accidentally upload extra files

.github/workflows/live-testing.yml
OhmV-IR, line 35: Copy of seggan's comment, this needs to use the gradle runServer task

OhmV-IR (changes requested):
Needs some changes

.github/workflows/gradle.yml
LordIdra: Just FYI jar will be bigger (maybe 20MB) once we shade in the database for blockstorage

.github/workflows/discord-webhook.yml
LordIdra: don't think it matters tbh, I think this is fine since it's already implemented and we won't have to change it much if at all

LordIdra:
> Fixed / Perfected most of the stuff, just kept some stuff "unresolved" so that other people can see and give their views

Any thoughts on the gradle runServer task thing? Also resolved the discussion about how long to keep jars because we have like 100 builds to go before we run into that problem (feel free to comment if you want to discuss it further though)

Seggan:
> > > Mostly everything is done, just need to do conversion of the code into `runServer` task, does that have an automatic timeout?
> > 
> > No. That can be fixed tho.
> 
> I did runServer thing, but just a thing to keep in mind is that if the action exceeds the timeout [3mins I kept it] it'll cause the action to "fail" even if the server runs, we just need to implement a automatic stop in the gradle script after like running the server in order to make it pass

Apparently Gradle tasks have a timeout parameter, maybe we could use that

.github/workflows/gradle.yml
```diff
@@ -2,48 +2,73 @@ name: Kotlin CI with Gradle
...
+
+      - name: Setup Gradle
+        uses: gradle/actions/setup-gradle@v4.2.2
 
-    # Configure Gradle for optimal use in GitHub Actions, including caching of downloaded dependencies.
-    # See: https://github.com/gradle/actions/blob/main/setup-gradle/README.md
-    - name: Setup Gradle
-      uses: gradle/actions/setup-gradle@af1da67850ed9a4cedd57bfd976089dd991e2582 # v4.0.0
+      - name: Build with Gradle Wrapper
+        run: ./gradlew build -Pversion=${{ inputs.version || github.run_number }}-SNAPSHOT
```
OhmV-IR, line 42: Should be able to get rid of `|| github.run_number` because you already set it as the default if inputs.version is not provided in L16

.github/workflows/gradle.yml
```diff
@@ -2,48 +2,73 @@ name: Kotlin CI with Gradle
...
 
-    - name: Build with Gradle Wrapper
-      run: ./gradlew build
+      - name: Upload the artifact
+        uses: actions/upload-artifact@v4.6.0
+        with:
+          name: pylon-core-${{ inputs.version || github.run_number }}
+          path: 'build/libs/pylon-core-${{ inputs.version || github.run_number }}.jar'
```
OhmV-IR, line 48: same here, should be able to get rid of `|| github.run_number`

.github/workflows/gradle.yml
```diff
@@ -2,48 +2,73 @@ name: Kotlin CI with Gradle
...
+          distribution: 'temurin'
+
+      - name: Generate and submit dependency graph
+        uses: gradle/actions/dependency-submission@v4.2.2
+
+  test-build:
+    needs: [build]
+    uses: ./.github/workflows/live-testing.yml
+    with:
+      artifact-name: pylon-core-${{ inputs.version || github.run_number }}
```
OhmV-IR, line 74: Can get rid of this as input is not needed since gradle runServer task builds the plugin anyways

.github/workflows/live-testing.yml
```diff
@@ -0,0 +1,41 @@
+name: Live Testing
+
+on:
+  workflow_call:
+    inputs:
+      artifact-name:
+        required: true
+        type: string
```
OhmV-IR, line 8: Because you don't need to download the build as the runServer gradle task builds it for you, this isn't needed

.github/workflows/live-testing.yml
```diff
@@ -0,0 +1,41 @@
...
+      - name: Setup server
+        run: |
+          echo 'eula=true' > eula.txt
+          mkdir plugins
+
+      - name: Download Pylon
+        uses: actions/download-artifact@v4.1.8
+        with:
+          name: ${{ inputs.artifact-name }}
+          path: plugins/
```
OhmV-IR, line 35: Plugin is built by the runServer task anyways, this isn't needed

OhmV-IR (changes requested):
Just two easy things to fix and then I'll LGTM this ASAP

Seggan (approved):

PR #3, opened 2025-01-17
Position class updates

LordIdra:
Changes to BlockPosition and ChunkPosition as discussed on discord

PR #4, opened 2025-01-17
Add abstract network

Seggan:
A basic network. To be merged after #3.


src/main/kotlin/io/github/pylonmc/pylon/core/network/Network.kt
```diff
@@ -0,0 +1,82 @@
...
+
+    val leaves: Set<NetworkTree.Leaf>
+        get() = allNodes.filterIsInstance<NetworkTree.Leaf>().toSet()
+
+    val branches: Set<NetworkTree.Branch>
+        get() = allNodes.filterIsInstance<NetworkTree.Branch>().toSet()
+
+    abstract fun isValidBlock(block: Block): Boolean
+
+    open fun scan() {
```
LordIdra, line 27: This is O(n), and if we need to do this constantly I can imagine it being costly. Maybe it would be better to listen for yet-to-be-implemented PylonBlockPlace and PylonBlockBreak events and update the network based on them?

src/main/kotlin/io/github/pylonmc/pylon/core/network/Network.kt
```diff
@@ -0,0 +1,82 @@
...
+
+    abstract val checkDistance: Int
+
+    open val maxNetworkDepth: Int = Int.MAX_VALUE
+
+    var root: NetworkTree = NetworkTree.Leaf(origin.block.position)
+    var allNodes: Set<NetworkTree> = setOf(root)
+        private set
+
+    val leaves: Set<NetworkTree.Leaf>
```
LordIdra, line 19: these two are expensive operations I think, should we just cache them when the tree is modified?

src/main/kotlin/io/github/pylonmc/pylon/core/network/Network.kt
```diff
@@ -0,0 +1,82 @@
...
+            return if (children.isEmpty()) {
+                NetworkTree.Leaf(block.position)
+            } else {
+                NetworkTree.Branch(block.position, children.map { it.toImmutable() }.toSet())
+            }
+        }
+    }
+}
+
+private val ORTHOGONAL = arrayOf(
```
LordIdra, line 75: Think we should make this a protected field so we can override if necessary?

src/main/kotlin/io/github/pylonmc/pylon/core/network/Network.kt
```diff
@@ -0,0 +1,82 @@
...
+        val rootTree = MutableNetworkTree(origin.block)
+        val scanQueue = ArrayDeque(listOf(NextScan(origin.block, 0, rootTree)))
+        val visited = mutableSetOf<BlockPosition>()
+        while (scanQueue.isNotEmpty()) {
+            val (block, currentDepth, parentNode) = scanQueue.removeFirst()
+            if (currentDepth > maxNetworkDepth || !visited.add(block.position)) {
+                continue
+            }
+            val currentNode = MutableNetworkTree(block)
+            parentNode.children.add(currentNode)
```
LordIdra, line 37: There doesn't seem to be a distinction between which blocks can and cannot have children if I'm reading this correctly?

LordIdra (changes requested):

src/main/kotlin/io/github/pylonmc/pylon/core/network/Network.kt
Seggan: No, there isn't yet

src/main/kotlin/io/github/pylonmc/pylon/core/network/Network.kt
Seggan: What if I want vanilla blocks (like the unimplemented thing with iron bars for electricity). What if I want blocks that are only a certain state?

LordIdra: Then we could listen to vanilla block events, but tbh I don't think we'll need to do that. Even if we couldn't, IMO it's still worth listening to these events because they would _massively_ reduce performance overhead for large networks (which as we learnt with slimefun is quite important)

Seggan: That can be left for implementations/consumers. This is only a base class, and `scan` can be called at any time, not just every tick or something.

LordIdra: I don't think that should be left for implementations, because it's fundamentally how the class builds the network. I don't see why we'd need scan in this case

LordIdra (approved):

Intybyte (approved):

PR #6, opened 2025-01-19
Pylon persistent data container

LordIdra:
Adds our implementation of PersistentDataContainer for use in BlockStorage.

I will add testing for this (including nested PDCs) in another PR when we have the test addon going.

### DIFF
--- a/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/PylonDataWriter.kt
+++ b/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/PylonDataWriter.kt
@@ -1,0 +1,8 @@
+package io.github.pylonmc.pylon.core.persistence
+
+import org.bukkit.NamespacedKey
+import org.bukkit.persistence.PersistentDataType
+
+interface PylonDataWriter {
+    fun <P : Any?, C : Any?> set(key: NamespacedKey, type: PersistentDataType<P, C>, value: C & Any)
+}
--- a/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/PylonBlock.kt
+++ b/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/PylonBlock.kt
@@ -1,17 +1,17 @@
 package io.github.pylonmc.pylon.core.block
 
+import io.github.pylonmc.pylon.core.persistence.PylonDataReader
+import io.github.pylonmc.pylon.core.persistence.PylonDataWriter
 import io.github.pylonmc.pylon.core.registry.PylonRegistries
 import io.github.pylonmc.pylon.core.registry.PyonRegistryKeys
-import io.github.pylonmc.pylon.core.state.StateReader
-import io.github.pylonmc.pylon.core.state.StateWriter
 import org.bukkit.NamespacedKey
 import org.bukkit.block.Block
 
-open class PylonBlock<S : PylonBlockSchema> private constructor(val schema: S, val block: Block) {
-    constructor(stateReader: StateReader, block: Block)
-            : this(getSchemaOfType<S>(stateReader.id), block)
+open class PylonBlock<S: PylonBlockSchema> private constructor(val schema: S, val block: Block) {
+    constructor(reader: PylonDataReader, block: Block)
+            : this(getSchemaOfType<S>(reader.id), block)
 
-    fun write(writer: StateWriter) {}
+    fun write(writer: PylonDataWriter) {}
 
     // TODO listener
     fun tick() {}
--- a/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/PylonBlockSchema.kt
+++ b/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/PylonBlockSchema.kt
@@ -1,8 +1,8 @@
 package io.github.pylonmc.pylon.core.block
 
+import io.github.pylonmc.pylon.core.persistence.PylonDataReader
 import io.github.pylonmc.pylon.core.registry.PylonRegistries
 import io.github.pylonmc.pylon.core.registry.PyonRegistryKeys
-import io.github.pylonmc.pylon.core.state.StateReader
 import org.bukkit.Keyed
 import org.bukkit.Material
 import org.bukkit.NamespacedKey
@@ -25,7 +25,7 @@ open class PylonBlockSchema(
     internal val loadConstructor: MethodHandle = try {
         MethodHandles.lookup().unreflectConstructor(
             blockClass.getConstructor(
-                StateReader::class.java,
+                PylonDataReader::class.java,
                 Block::class.java
             )
         )
--- a/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/PylonDataReader.kt
+++ b/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/PylonDataReader.kt
@@ -4,7 +4,9 @@ import org.bukkit.NamespacedKey
 import org.bukkit.persistence.PersistentDataType
 
 interface PylonDataReader {
-    // ID of the holder of this reader - for example the ID of a block
+    /**
+      * ID of the holder of this reader - for example the ID of a block
+      */
     val id: NamespacedKey
 
     fun <P : Any, C : Any> has(key: NamespacedKey, type: PersistentDataType<P, C>): Boolean
--- a/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/PylonDataReader.kt
+++ b/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/PylonDataReader.kt
@@ -1,0 +1,25 @@
+package io.github.pylonmc.pylon.core.persistence
+
+import org.bukkit.NamespacedKey
+import org.bukkit.persistence.PersistentDataType
+
+interface PylonDataReader {
+    // ID of the holder of this reader - for example the ID of a block
+    val id: NamespacedKey
+
+    fun <P : Any, C : Any> has(key: NamespacedKey, type: PersistentDataType<P, C>): Boolean
+
+    fun has(key: NamespacedKey): Boolean
+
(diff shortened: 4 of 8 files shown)


src/main/kotlin/io/github/pylonmc/pylon/core/persistence/PylonPersistentDataContainer.kt
```diff
@@ -0,0 +1,183 @@
...
+                    val bytes = ByteBuffer.allocate(intArray.size * Long.SIZE_BYTES)
+                    for (x in intArray) {
+                        bytes.putLong(x)
+                    }
+                    bytes.array()
+                }
+                // This is PylonPersistentDataContainer here because when we get to deserializing PDCs, there is no way to know
+                // which implementor of PDC to serialize to. Therefore, we will only allow Pylon PDCs to be nested.
+                PersistentDataContainer::class.java -> (primitive as PylonPersistentDataContainer).serializeToBytes()
+                else -> error("Unrecognized primitive type")
```
Seggan, line 147:
```suggestion
                else -> error("Unrecognized primitive type $primitiveType")
```

src/main/kotlin/io/github/pylonmc/pylon/core/persistence/PylonPersistentDataContainer.kt
```diff
@@ -0,0 +1,183 @@
...
+            }
+
+        // Sadly no way to 'switch' on the type T to avoid unchecked casting
+        // See https://stackoverflow.com/questions/73523156/how-to-overload-function-with-different-return-types-and-the-same-parameters-in
+        @Suppress("UNCHECKED_CAST")
+        private fun <T: Any?> bytesToPrimitive(primitiveType: Class<T>, bytes: ByteArray): T
```
Seggan, line 153: I think all the `as T` here can be replaced with `primitiveTypr.cast(...)`. At least get runtime verification.

src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
```diff
@@ -0,0 +1,60 @@
+package io.github.pylonmc.pylon.core.persistence
+
+import org.bukkit.NamespacedKey
+import org.bukkit.persistence.PersistentDataAdapterContext
+import org.bukkit.persistence.PersistentDataType
+import java.nio.ByteBuffer
+import java.util.*
+
+object Serializers {
```
Seggan, line 9: I think these can get their own files and subpackage

Seggan (changes requested):
I prefer putting helper functions at top level instead of companion object but thats just personal preference, I won't block this pr because of that.

src/main/kotlin/io/github/pylonmc/pylon/core/persistence/PylonPersistentDataContainer.kt
```diff
@@ -0,0 +1,183 @@
+package io.github.pylonmc.pylon.core.persistence
+
+import io.github.pylonmc.pylon.core.pluginInstance
+import org.bukkit.NamespacedKey
+import org.bukkit.persistence.PersistentDataAdapterContext
+import org.bukkit.persistence.PersistentDataContainer
+import org.bukkit.persistence.PersistentDataType
+import java.nio.ByteBuffer
+
+/*
```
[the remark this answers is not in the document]
LordIdra: oh yeah thanks lol

src/main/kotlin/io/github/pylonmc/pylon/core/persistence/PylonPersistentDataContainer.kt
LordIdra: Also added a new exception for this

src/main/kotlin/io/github/pylonmc/pylon/core/persistence/PylonPersistentDataContainer.kt
LordIdra: Nope still have to have as, like `primitiveType.cast(...) as T` because kotlin detects what the `when` block is doing and knows that, for example, if I use .cast in the short block, the return of cast is to a short (which then has to be cast to T). and the return type has to be T. Would cast still add more safety in this case?

src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
LordIdra: Idk seems fine to have them here IMO, any other opinions?

src/main/kotlin/io/github/pylonmc/pylon/core/persistence/PylonPersistentDataContainer.kt
Seggan: Never mind then

Intybyte:
Any reson why you have a bunch of <Generic : Any> ? You can just drop the Any it is like saying <T extends Object> which doesnt make much sense

src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
OhmV-IR: uh considering I have a branch that has already diverged with this set of changes, I'd rather not goof up my branch history any further

LordIdra:
> Any reson why you have a bunch of ? You can just drop the Any it is like saying which doesnt make much sense

in the method definitions yes because I'm implementing an interface, if I don't specify ? then kotlin complains I'm not implementing the method

src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
LordIdra: Ok gonna just resolve for now then

src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
```diff
@@ -0,0 +1,60 @@
...
+    val DOUBLE = PersistentDataType.DOUBLE!!
+    val BOOLEAN = PersistentDataType.BOOLEAN!!
+    val STRING = PersistentDataType.STRING!!
+    val BYTE_ARRAY = PersistentDataType.BYTE_ARRAY!!
+    val INTEGER_ARRAY = PersistentDataType.INTEGER_ARRAY!!
+    val LONG_ARRAY = PersistentDataType.LONG_ARRAY!!
+    val TAG_CONTAINER = PersistentDataType.TAG_CONTAINER!!
+    val LIST = PersistentDataType.LIST!!
+    val NAMESPACED_KEY = NamespacedKeyPersistentDataType()
+}
```
OhmV-IR, line 24: I think you forgot to add UUID here with the other persistent data types

OhmV-IR (changes requested):
LGTM CW except for one change, easy one though

src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
```diff
@@ -0,0 +1,61 @@
...
+        = complex.toString().toByteArray(Charsets.UTF_8)
+}
+
+class UUIDPersistentDataType : PersistentDataType<ByteArray, UUID> {
+    override fun getPrimitiveType(): Class<ByteArray>
+            = ByteArray::class.java
+
+    override fun getComplexType(): Class<UUID>
+            = UUID::class.java
+
```
Seggan, line 47: Why not LongArray instead

Seggan (changes requested):

src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
LordIdra: wdym?

LordIdra: oh wait I assume you mean primitive type?

src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
Intybyte: I think we should open an issue/pr to solve this because having 3 classes in a single file is very unsettling to me, i can turn a blind eye to the Excpetions file due to its simplicity

src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
LordIdra: Ok wait actually I know why - we might want to reuse it in future, which is easier if it's bytes

Seggan (approved):
Not really happy with "using certain primitives because composability" but approving anyway

Intybyte (approved):

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/PylonDataWriter.kt
```diff
@@ -0,0 +1,8 @@
+package io.github.pylonmc.pylon.core.persistence
+
+import org.bukkit.NamespacedKey
+import org.bukkit.persistence.PersistentDataType
+
+interface PylonDataWriter {
+    fun <P : Any?, C : Any?> set(key: NamespacedKey, type: PersistentDataType<P, C>, value: C & Any)
```
Seggan, line 7: Why not 

```suggestion
    fun <P : Any?, C : Any> set(key: NamespacedKey, type: PersistentDataType<P, C>, value: C)
```

Seggan (changes requested):

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/PylonDataWriter.kt
LordIdra: Because I'm implementing a Java interface (PersistentDataContainer) and this is how the method was translated. If I implement your suggestion I get an error because it thinks I'm not implementing PylonDataWriter because the function signature is slightly different

Seggan: Weird but ok

Seggan (approved):

PR #7, opened 2025-01-20
Add serializers for more datatypes

OhmV-IR:
This PR adds serializers for:
- Vector
- World
- Location
- BlockPosition
- ChunkPosition
- Char
- No longer doing KV map, idra is taking it over, out of my skill level
Rebased with merge of test-addon, have a framework for implementing tests here but the actual tests will have to be added in a future PR after there is a way to create a PersistentDataAdapterContext reference
Code review should be pretty thorough, this is basically my first run at kotlin
Waiting on #6 to be merged before this


pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
```diff
@@ -0,0 +1,282 @@
...
+    }
+
+    override fun toPrimitive(complex: Map<*, *>, context: PersistentDataAdapterContext): ByteArray {
+        val bytesOut = ByteArrayOutputStream()
+        val outputStream = ObjectOutputStream(bytesOut)
+        outputStream.writeObject(complex)
+        outputStream.flush()
+        return bytesOut.toByteArray()
+    }
+}
```
Seggan, line 282: I can't help but this this can be done better using nested `PersistentDataContainer`s

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
```diff
@@ -0,0 +1,282 @@
...
+    }
+
+    override fun toPrimitive(complex: ChunkPosition, context: PersistentDataAdapterContext): ByteArray {
+        if(complex.world != null){
+            val buffer = ByteBuffer.allocate(2 * Long.SIZE_BYTES + 2 * Int.SIZE_BYTES)
+            buffer.putLong(complex.world!!.uid.mostSignificantBits)
+            buffer.putLong(complex.world!!.uid.leastSignificantBits)
+            buffer.putInt(complex.x)
+            buffer.putInt(complex.z)
+            complex.asLong
```
Seggan, line 231: ??

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
```diff
@@ -0,0 +1,282 @@
...
+            return buffer.array()
+        }
+        else{
+            val buffer = ByteBuffer.allocate(2 * Int.SIZE_BYTES)
+            buffer.putInt(complex.x)
+            buffer.putInt(complex.z)
+            return buffer.array()
+        }
```
Seggan, line 239:
```suggestion
        val buffer = ByteBuffer.allocate(2 * Long.SIZE_BYTES + 2 * Int.SIZE_BYTES)
        val world = complex.world
        if (world != null) {
            buffer.putLong(world.uid.mostSignificantBits)
            buffer.putLong(world.uid.leastSignificantBits)
        }
        buffer.putInt(complex.x)
        buffer.putInt(complex.z)
        return buffer.array()
```
This can be massively simplified

pylon-test/src/main/java/io/github/pylonmc/pylon/test/TestSerializers.kt
```diff
@@ -0,0 +1,21 @@
...
+        fun testAllSerializers(): Boolean {
+            try {
+                for (method in this.javaClass.methods) {
+                    if (method.name != "TestAllSerializers" && !(method.invoke(null, null) as Boolean)) {
+                        return false
+                    }
+                }
+                return true
+            }
+            catch(e: Exception){
```
Seggan, line 15:
```suggestion
            } catch (e: Exception) {
```
Formatting

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
```diff
@@ -0,0 +1,282 @@
...
+            val x = buffer.getInt()
+            val z = buffer.getInt()
+            return ChunkPosition(Bukkit.getServer().getWorld(UUID(mostSignificantBits, leastSignificantBits)), x, z)
+        }
+        else{
+            val buffer = ByteBuffer.wrap(primitive)
+            val x = buffer.getInt()
+            val z = buffer.getInt()
+            return ChunkPosition(null, x, z)
+        }
```
Seggan, line 221:
```suggestion
        val buffer = ByteBuffer.wrap(primitive)
        val uid = if (primitive.size > 2 * Int.SIZE_BYTES) {
            val mostSignificantBits = buffer.getLong()
            val leastSignificantBits = buffer.getLong()
            UUID(mostSignificantBits, leastSignificantBits)
        } else {
            null
        }
        val x = buffer.getInt()
        val z = buffer.getInt()
        return ChunkPosition(Bukkit.getWorld(uid), x, z)
```
Can be simplified

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
```diff
@@ -0,0 +1,282 @@
...
+            buffer.putInt(complex.z)
+            return buffer.array()
+        }
+        else{
+            val buffer = ByteBuffer.allocate(3 * Int.SIZE_BYTES)
+            buffer.putInt(complex.x)
+            buffer.putInt(complex.y)
+            buffer.putInt(complex.z)
+            return buffer.array()
+        }
```
Seggan, line 196: These can be massively simplified, see below

Seggan (changes requested):
A C# lover I see

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
```diff
@@ -0,0 +1,282 @@
...
+
+class BlockPositionPersistentDataType : PersistentDataType<ByteArray, BlockPosition> {
+    override fun getPrimitiveType(): Class<ByteArray>
+        = ByteArray::class.java
+
+    override fun getComplexType(): Class<BlockPosition>
+        = BlockPosition::class.java
+
+    override fun fromPrimitive(primitive: ByteArray, context: PersistentDataAdapterContext): BlockPosition {
+        if(primitive.size > 3 * Int.SIZE_BYTES){
```
LordIdra, line 162: Just a small thing but by convention in this project we're formatting if/else like
if (something) {
    ...
} else {
    ...
}

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
```diff
@@ -0,0 +1,282 @@
...
+            val buffer = ByteBuffer.allocate(3 * Int.SIZE_BYTES)
+            buffer.putInt(complex.x)
+            buffer.putInt(complex.y)
+            buffer.putInt(complex.z)
+            return buffer.array()
+        }
+    }
+}
+
+class ChunkPositionPersistentDataType : PersistentDataType<ByteArray, ChunkPosition> {
```
LordIdra, line 200: Exact same things here as for BlockPosition

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
```diff
@@ -0,0 +1,282 @@
...
+    }
+
+    override fun toPrimitive(complex: Char, context: PersistentDataAdapterContext): ByteArray {
+        val buffer = ByteBuffer.allocate(Char.SIZE_BYTES)
+        buffer.putChar(complex)
+        return buffer.array()
+    }
+}
+
+class MapPersistentDataType : PersistentDataType<ByteArray, Map<*, *>> {
```
LordIdra, line 262: Ok so for this we don't want to do Map<*, *> because that's saying 'allow a map from anything to anything - _even datatypes that we don't necessarily know can be serialized._' So for example I could feed in a Map<String, Set<PylonMachine>> and even though we can't actually serialize that, this function will accept it and just run with it. 

Well, ok, technically it _can_ be serialized as you're doing here with writeObject, but this is extremely unstable. For example imagine I have a class with some field x which is an integer. Then, I save some data. I then change that field to be a string. Now, this will error and throw an exception whenever it tries to deserialize. Also, what if my object has a reference to another object? Now we can't deserialize it because that reference might no longer exist. And so on. But you will not actually get an error about this with the current implementation - it'll work for some types but mysteriously fail at different points in different ways. This is the point in writing serializers like we're doing here

What you want to do instead is add type parameters and make sure the type parameters are PersistentDataTypes. That guarantees that the inputs are safe to serialize. You can look at PersistentDataType.LIST / ListPersistentDataType to see how this is implemented for lists, it's very similar to what you'll need to do here.

LordIdra (changes requested):

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
LordIdra: This should be implemented differently (see my comment)

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
```diff
@@ -0,0 +1,282 @@
...
+        buffer.putLong(complex.world.uid.mostSignificantBits)
+        buffer.putLong(complex.world.uid.leastSignificantBits)
+        buffer.putDouble(complex.x)
+        buffer.putDouble(complex.y)
+        buffer.putDouble(complex.z)
+        return buffer.array()
+    }
+}
+
+class BlockPositionPersistentDataType : PersistentDataType<ByteArray, BlockPosition> {
```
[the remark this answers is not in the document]
Seggan: Hint: `LongArray` is an allowed primitive type

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
LordIdra: Yeah I suggested using a byte to indicate whether the world is present in mine but this would also work. Up to you ohm

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
LordIdra: As previously - yeah I suggested using a byte to indicate whether the world is present in mine but this would also work. Up to you ohm

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
LordIdra: (same as on the above 2 comments)

Seggan:
Well that didn't work

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
OhmV-IR: yeah I don't even know LMAO, fixed will push soon

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
OhmV-IR: eh I'd rather not, this is the part of the code that kinda needs to get hyperoptimized

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
```diff
@@ -0,0 +1,282 @@
...
+    override fun toPrimitive(complex: Vector, context: PersistentDataAdapterContext): ByteArray {
+        val buffer = ByteBuffer.allocate(3 * Double.SIZE_BYTES)
+        buffer.putDouble(complex.x)
+        buffer.putDouble(complex.y)
+        buffer.putDouble(complex.z)
+        return buffer.array()
+    }
+}
+
+class WorldPersistentDataType : PersistentDataType<ByteArray, World> {
```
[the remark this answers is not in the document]
OhmV-IR: done

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
```diff
@@ -0,0 +1,282 @@
...
+
+    override fun toPrimitive(complex: World, context: PersistentDataAdapterContext): ByteArray {
+        val buffer = ByteBuffer.allocate(2 * Long.SIZE_BYTES)
+        buffer.putLong(complex.uid.mostSignificantBits)
+        buffer.putLong(complex.uid.leastSignificantBits)
+        return buffer.array()
+    }
+}
+
+class LocationPersistentDataType : PersistentDataType<ByteArray, Location> {
```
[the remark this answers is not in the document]
OhmV-IR: done

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
OhmV-IR: no more elses in here now but will remember this

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
```diff
@@ -0,0 +1,282 @@
...
+        = ByteArray::class.java
+
+    override fun getComplexType(): Class<World>
+        = World::class.java
+
+    override fun fromPrimitive(primitive: ByteArray, context: PersistentDataAdapterContext): World {
+        val buffer = ByteBuffer.wrap(primitive)
+        val mostSignificantBits = buffer.getLong()
+        val leastSignificantBits = buffer.getLong()
+        return Bukkit.getServer().getWorld(UUID(mostSignificantBits, leastSignificantBits))!!
```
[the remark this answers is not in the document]
OhmV-IR: done

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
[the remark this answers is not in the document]
OhmV-IR: done

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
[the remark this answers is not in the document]
OhmV-IR: Actually if I use asLong to initialize it, for some reason the world param is no longer nullable so if I don't have a world attached to the blockposition and I only have the asLong in the fromPrimitive, I need to somehow deconstruct that into xyz coords to be able to reconstruct the object.

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
OhmV-IR: Same problem with ChunkPosition if you use asLong to initialize the object, the world param is no longer nullable

OhmV-IR: implemented first byte to indicate world presence

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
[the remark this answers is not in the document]
OhmV-IR: implemented

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
```diff
@@ -0,0 +1,288 @@
...
+
+    override fun toPrimitive(complex: Char, context: PersistentDataAdapterContext): ByteArray {
+        val buffer = ByteBuffer.allocate(Char.SIZE_BYTES)
+        buffer.putChar(complex)
+        return buffer.array()
+    }
+}
+
+object MapSerializers {
+    val UUID_FLOAT = MapPersistentDataTypeUUIDFloat()
```
LordIdra, line 237: you don't have to have different implementations for each type of map - you can just have one generic map which takes K and V PersistentDataType parameters, and use K and V's serializers, or something like that

LordIdra:
What exactly is the problem with the PDC context?

OhmV-IR:
Oh no real problem just that there's no way to actually use the serializers externally because there's no way to create a PersistentDataContainer which is required for the PersistentDataAdapterContext so tests for these serializers will have to wait until after there is a public constructor for it

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/Exceptions.kt
```diff
@@ -5,3 +5,4 @@ class NotRegisteredException(id: String) : Exception("'$id' is not registered")
 class AlreadyRegisteredException(id: String) : Exception("'$id' is already registered")
 class MissingPlaceConstructorException(id: String) : Exception("'$id' is missing a place constructor")
 class MissingLoadConstructorException(id: String) : Exception("'$id' is missing a load constructor")
+class InvalidWorldUidException(uid: String) : Exception("Deserialized world uid '$uid' was not a valid world")
```
Seggan, line 8: Just use an IllegalArgumentException, no need for a custom one

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
```diff
@@ -0,0 +1,288 @@
...
+}
+
+class VectorPersistentDataType : PersistentDataType<ByteArray, Vector> {
+    override fun getPrimitiveType(): Class<ByteArray>
+        = ByteArray::class.java
+
+    override fun getComplexType(): Class<Vector>
+        = Vector::class.java
+
+    override fun fromPrimitive(primitive: ByteArray, context: PersistentDataAdapterContext): Vector {
```
Seggan, line 85: IIRC DoubleArray is an allowed primitive type in PDC, just use that instead of doing this byte buffer stuff

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
```diff
@@ -0,0 +1,288 @@
...
+    override fun getPrimitiveType(): Class<ByteArray>
+        = ByteArray::class.java
+
+    override fun getComplexType(): Class<ChunkPosition>
+        = ChunkPosition::class.java
+
+    override fun fromPrimitive(primitive: ByteArray, context: PersistentDataAdapterContext): ChunkPosition {
+        val buffer = ByteBuffer.wrap(primitive)
+        var world: World? = null
+        if(buffer.get() == 1.toByte()){
```
Seggan, line 194: If it was serialized with no world, this would cause data corruption. You need to add a 0 or smth to pad in the case of no world

Seggan (changes requested):

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/Exceptions.kt
OhmV-IR: done

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
OhmV-IR: done

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
OhmV-IR: done

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
```diff
@@ -0,0 +1,229 @@
...
+    val VECTOR = VectorPersistentDataType()
+    val WORLD = WorldPersistentDataType()
+    val BLOCK_POSITION = BlockPositionPersistentDataType()
+    val CHUNK_POSITION = ChunkPositionPersistentDataType()
+    val LOCATION = LocationPersistentDataType()
+    val CHAR = CharPersistentDataType()
+//    val MAP = MapPersistentDataType()
+}
+
+class NamespacedKeyPersistentDataType : PersistentDataType<ByteArray, NamespacedKey> {
```
Seggan, line 42: This can use `String` as primitive

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
```diff
@@ -0,0 +1,229 @@
...
+        = NamespacedKey::class.java
+
+    override fun fromPrimitive(primitive: ByteArray, context: PersistentDataAdapterContext): NamespacedKey
+        = NamespacedKey.fromString(primitive.toString(Charsets.UTF_8))!!
+
+    override fun toPrimitive(complex: NamespacedKey, context: PersistentDataAdapterContext): ByteArray
+        = complex.toString().toByteArray(Charsets.UTF_8)
+}
+
+class UUIDPersistentDataType : PersistentDataType<ByteArray, UUID> {
```
Seggan, line 56: This can use `LongArray` as primitve

Seggan (changes requested):
Sorry, missed these two last review

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
LordIdra: I don't think it should, so that we can reuse this serializer later if we need to store a UUID for some reason (as mentioned in PDC PR)

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
OhmV-IR: done

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
OhmV-IR: reverted because other ones depend on this and are using bytearray

Seggan (approved):
Yeah just what Idra said, otherwise LGTM

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
```diff
@@ -0,0 +1,225 @@
...
+        return BlockPosition(world, x, y, z)
+    }
+
+    override fun toPrimitive(complex: BlockPosition, context: PersistentDataAdapterContext): ByteArray {
+        val buffer = ByteBuffer.allocate(2 * Long.SIZE_BYTES + 3 * Int.SIZE_BYTES + 1)
+        val world = complex.world
+        if(world != null){
+            buffer.put(1)
+            buffer.put(Serializers.WORLD.toPrimitive(complex.world!!, context))
+        }
```
Intybyte, line 165: Can't we just use the "nil" UUID (all zeros) to check this instead of adding extra data?

pylon-test/src/main/java/io/github/pylonmc/pylon/test/TestAddon.java
```diff
@@ -10,7 +10,12 @@ public class TestAddon extends JavaPlugin implements PylonAddon {
     @Override
     public void onEnable() {
         getLogger().info("Test addon enabled!");
-
+        if(TestSerializers.testAllSerializers()){
+            getLogger().info("All serializer tests passed!");
+        }
+        else{
+            getLogger().severe("At least one serializer test failed!");
```
Intybyte, line 17: identation please

Intybyte (changes requested):

pylon-test/src/main/java/io/github/pylonmc/pylon/test/TestAddon.java
OhmV-IR: fixed to align with `} else {` format idra mentioned earlier

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
OhmV-IR: could I suppose but then we have to check 2 longs instead of 1 byte and 1 byte isn't too expensive in storage space terms

OhmV-IR:
Need full CW review again, entirely possible I goofed the merge(s) / rebase(s)

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
```diff
@@ -67,3 +86,155 @@ class UUIDPersistentDataType : PersistentDataType<LongArray, UUID> {
...
+        buffer.put(0)
```
Seggan, line 181:
```suggestion
        if(world != null){
            buffer.put(1)
            buffer.put(Serializers.WORLD.toPrimitive(complex.world!!, context))
        } else {
            buffer.put(0)
        }
```
Data corruption possibility

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
```diff
@@ -67,3 +86,155 @@ class UUIDPersistentDataType : PersistentDataType<LongArray, UUID> {
...
+    }
+
+    override fun toPrimitive(complex: ChunkPosition, context: PersistentDataAdapterContext): ByteArray {
+        val buffer = ByteBuffer.allocate(2 * Long.SIZE_BYTES + 2 * Int.SIZE_BYTES + 1)
+        val world = complex.world
+        if (world != null) {
+            buffer.put(1)
+            buffer.put(Serializers.WORLD.toPrimitive(complex.world!!, context))
+        }
+        buffer.put(0)
```
Seggan, line 216:
```suggestion
        if (world != null) {
            buffer.put(1)
            buffer.put(Serializers.WORLD.toPrimitive(complex.world!!, context))
        } else {
            buffer.put(0)
        }
```
Same as above

Seggan (changes requested):

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/Serializers.kt
OhmV-IR: whoops yeah idk how I managed to goof that up

Seggan (approved):

LordIdra (changes requested):

pylon-test/build.gradle.kts
LordIdra, line 1: what happened here? why'd this get yeeted

pylon-test/src/main/java/io/github/pylonmc/pylon/test/TestAddon.java
LordIdra, line 1: and why's this yeeted

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/PylonSerializers.kt
```diff
@@ -0,0 +1,248 @@
...
+        = NamespacedKey::class.java
+
+    override fun fromPrimitive(primitive: String, context: PersistentDataAdapterContext): NamespacedKey
+        = NamespacedKey.fromString(primitive)!!
+
+    override fun toPrimitive(complex: NamespacedKey, context: PersistentDataAdapterContext): String
+        = complex.toString()
+}
+
+class UUIDPersistentDataType : PersistentDataType<ByteArray, UUID> {
```
LordIdra, line 74: This is back to ByteArray, is that intentional?

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/PylonSerializers.kt
```diff
@@ -0,0 +1,248 @@
...
+        val uid = Serializers.UUID.fromPrimitive(primitive, context)
+        return Bukkit.getWorld(uid) ?: throw IllegalArgumentException(uid.toString())
+    }
+
+    override fun toPrimitive(complex: World, context: PersistentDataAdapterContext): ByteArray {
+        return Serializers.UUID.toPrimitive(complex.uid, context)
+    }
+}
+
+class LocationPersistentDataType : PersistentDataType<ByteArray, Location> {
```
LordIdra, line 129: This could be a PersistentDataContainer as mentioned on Discord

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/PylonSerializers.kt
```diff
@@ -0,0 +1,248 @@
...
+        buffer.putDouble(complex.x)
+        buffer.putDouble(complex.y)
+        buffer.putDouble(complex.z)
+        buffer.putFloat(complex.yaw)
+        buffer.putFloat(complex.pitch)
+        return buffer.array()
+    }
+}
+
+class BlockPositionPersistentDataType : PersistentDataType<ByteArray, BlockPosition> {
```
LordIdra, line 159: same here, can be a PDC rather than ByteArray

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/PylonSerializers.kt
```diff
@@ -0,0 +1,248 @@
...
+            buffer.put(0)
+        }
+        buffer.putInt(complex.x)
+        buffer.putInt(complex.y)
+        buffer.putInt(complex.z)
+        return buffer.array()
+    }
+}
+
+class ChunkPositionPersistentDataType : PersistentDataType<ByteArray, ChunkPosition> {
```
LordIdra, line 196: and here (PDC instead of bytearray)

LordIdra (changes requested):

pylon-test/src/main/java/io/github/pylonmc/pylon/test/TestAddon.java
OhmV-IR: undone

pylon-test/build.gradle.kts
OhmV-IR: undone

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/PylonSerializers.kt
OhmV-IR: yes, a lot of other serializers use UUID, and since they use bytearray primarily they need the UUID serializer response to be in bytes as well

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/PylonSerializers.kt
OhmV-IR: done

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/PylonSerializers.kt
OhmV-IR: done

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/PylonSerializers.kt
OhmV-IR: done

pylon-test/src/main/java/io/github/pylonmc/pylon/test/TestAddon.java
LordIdra, line 1: could we just roll back all changes here to avoid rebasing the gametest pr?

pylon-test/src/main/java/io/github/pylonmc/pylon/test/TestSerializers.java
```diff
@@ -0,0 +1,36 @@
+package io.github.pylonmc.pylon.test;
```
LordIdra, line 1: this file can go - we can add testing once the test framework is in place

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/PylonSerializers.kt
```diff
@@ -0,0 +1,258 @@
...
+    override fun getPrimitiveType(): Class<PersistentDataContainer>
+            = PersistentDataContainer::class.java
+
+    override fun getComplexType(): Class<ChunkPosition>
+            = ChunkPosition::class.java
+
+    override fun fromPrimitive(primitive: PersistentDataContainer, context: PersistentDataAdapterContext): ChunkPosition {
+        val x = primitive.get(xKey, PersistentDataType.INTEGER)!!
+        val z = primitive.get(zKey, PersistentDataType.INTEGER)!!
+        if(primitive.has(worldKey)){
```
LordIdra, line 219: I don't believe you need this check, because `get` returns nullable so that would just get passed straight into the ChunkPosition constructor

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/PylonSerializers.kt
```diff
@@ -0,0 +1,258 @@
...
+            = PersistentDataContainer::class.java
+
+    override fun getComplexType(): Class<BlockPosition>
+            = BlockPosition::class.java
+
+    override fun fromPrimitive(primitive: PersistentDataContainer, context: PersistentDataAdapterContext): BlockPosition {
+        val x = primitive.get(xKey, PersistentDataType.INTEGER)!!
+        val y = primitive.get(yKey, PersistentDataType.INTEGER)!!
+        val z = primitive.get(zKey, PersistentDataType.INTEGER)!!
+        if(primitive.has(worldKey)){
```
LordIdra, line 185: same here with regards to get returning a nullable so no need for a has check

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/PylonSerializers.kt
```diff
@@ -0,0 +1,258 @@
...
+        val y = primitive.get(yKey, PersistentDataType.DOUBLE)!!
+        val z = primitive.get(zKey, PersistentDataType.DOUBLE)!!
+        val yaw = primitive.get(yawKey, PersistentDataType.FLOAT)!!
+        val pitch = primitive.get(pitchKey, PersistentDataType.FLOAT)!!
+        val world = Serializers.WORLD.fromPrimitive(primitive.get(worldKey, PersistentDataType.BYTE_ARRAY)!!, context)
+        return Location(world, x, y, z, yaw, pitch)
+    }
+
+    override fun toPrimitive(complex: Location, context: PersistentDataAdapterContext): PersistentDataContainer {
+        val PDC = context.newPersistentDataContainer()
```
LordIdra, line 157: PDC should be lowercase by convention

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/PylonSerializers.kt
LordIdra: they now use PDC though so this can go back to being a long array I think?

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/PylonSerializers.kt
```diff
@@ -0,0 +1,258 @@
...
+            return ChunkPosition(world, x, z)
+        }
+        return ChunkPosition(null, x, z)
+    }
+
+    override fun toPrimitive(complex: ChunkPosition, context: PersistentDataAdapterContext): PersistentDataContainer {
+        val PDC = context.newPersistentDataContainer()
+        PDC.set(xKey, PersistentDataType.INTEGER, complex.x)
+        PDC.set(zKey, PersistentDataType.INTEGER, complex.z)
+        if(complex.world != null){
```
LordIdra, line 230: Same here with nullability

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/PylonSerializers.kt
```diff
@@ -0,0 +1,254 @@
...
+    override fun getComplexType(): Class<Location>
+            = Location::class.java
+
+    override fun fromPrimitive(primitive: PersistentDataContainer, context: PersistentDataAdapterContext): Location {
+        val x = primitive.get(xKey, PersistentDataType.DOUBLE)!!
+        val y = primitive.get(yKey, PersistentDataType.DOUBLE)!!
+        val z = primitive.get(zKey, PersistentDataType.DOUBLE)!!
+        val yaw = primitive.get(yawKey, PersistentDataType.FLOAT)!!
+        val pitch = primitive.get(pitchKey, PersistentDataType.FLOAT)!!
+        val world = PylonSerializers.WORLD.fromPrimitive(primitive.get(worldKey, PersistentDataType.BYTE_ARRAY)!!, context)
```
LordIdra, line 152: What's going on here? Why are we storing world as a byte array? and then using WORLD.fromPrimitive?

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/PylonSerializers.kt
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+    override fun toPrimitive(complex: Location, context: PersistentDataAdapterContext): PersistentDataContainer {
+        val PDC = context.newPersistentDataContainer()
+        PDC.set(NamespacedKey.fromString("world")!!, PersistentDataType.BYTE_ARRAY, PylonSerializers.WORLD.toPrimitive(complex.world, context))
+        PDC.set(NamespacedKey.fromString("x")!!, PersistentDataType.DOUBLE, complex.x)
+        PDC.set(NamespacedKey.fromString("y")!!, PersistentDataType.DOUBLE, complex.y)
+        PDC.set(NamespacedKey.fromString("z")!!, PersistentDataType.DOUBLE, complex.z)
+        PDC.set(NamespacedKey.fromString("yaw")!!, PersistentDataType.FLOAT, complex.yaw)
+        PDC.set(NamespacedKey.fromString("pitch")!!, PersistentDataType.FLOAT, complex.pitch)
```
LordIdra, line 163: NamespaceKeys should be stored statically, rather than dynamically generated like they are here

LordIdra (changes requested):

pylon-test/src/main/java/io/github/pylonmc/pylon/test/TestAddon.java
OhmV-IR: done

pylon-test/src/main/java/io/github/pylonmc/pylon/test/TestSerializers.java
OhmV-IR: done

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/PylonSerializers.kt
OhmV-IR: done

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/PylonSerializers.kt
OhmV-IR: well since we're passing it through the world serializer which expects the input to be non-null we have to null check it anyways

pylon-test/build.gradle.kts
```diff
@@ -33,7 +33,9 @@ bukkit {
     apiVersion = "1.21"
     depend = listOf("pylon-core")
 }
-
+tasks.compileJava {
+    dependsOn(":pylon-core:jar")
+}
```
LordIdra: we don't need this task because of pylon-core:shadow-jar - also I don't think this would shade in dependencies anyway

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/PylonSerializers.kt
LordIdra: You shouldn't need to if you use `primitive.get(... PylonSerializers.WORLD)` and the same for set

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/PylonSerializers.kt
OhmV-IR: changed to long array since that is not the primitive type for world data serializer

LordIdra (approved):

PR #12, opened 2025-01-23
Map and set types

LordIdra:
Add map and set PylonPersistentDataType

### DIFF
--- a/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/PylonSerializers.kt
+++ b/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/PylonSerializers.kt
@@ -58,9 +58,9 @@ object PylonSerializers {
     @JvmField
     val CHAR = CharPersistentDataType()
     @JvmField
-    val SET = SetPersistentDataTypeProvider()
+    val SET = SetPersistentDataType.Companion
     @JvmField
-    val MAP = MapPersistentDataTypeProvider()
+    val MAP = MapPersistentDataType.Companion
 }
 
 class NamespacedKeyPersistentDataType : PersistentDataType<String, NamespacedKey> {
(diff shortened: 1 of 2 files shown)


pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/PylonSerializers.kt
```diff
@@ -240,4 +244,95 @@ class CharPersistentDataType : PersistentDataType<ByteArray, Char> {
...
+
+        override fun elementType(): PersistentDataType<P, C>
+                = elementType
+
+        override fun getPrimitiveType(): Class<PersistentDataContainer>
+                = PersistentDataContainer::class.java
+
+        @Suppress("UNCHECKED_CAST") // Yes, this is cursed, no, there's no way around it afaik
+        override fun getComplexType(): Class<Set<C>>
+                = Set::class.java as Class<Set<C>>
```
OhmV-IR, line 273: actually you can, just have to create an instance of it like this: `setOf<C>().javaClass`

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/persistence/PylonSerializers.kt
```diff
@@ -240,4 +244,95 @@ class CharPersistentDataType : PersistentDataType<ByteArray, Char> {
...
+
+        override fun valueType(): PersistentDataType<VP, VC>
+                = valueType
+
+        override fun getPrimitiveType(): Class<PersistentDataContainer>
+                = PersistentDataContainer::class.java
+
+        @Suppress("UNCHECKED_CAST")
+        override fun getComplexType(): Class<Map<KC, VC>>
+                = Map::class.java as Class<Map<KC, VC>>
```
OhmV-IR, line 317: Here you can use `arrayListOf<KC>().zip(arrayListOf<VC>()).toMap().javaClass` to be able to do this without an unchecked cast

OhmV-IR (changes requested):
Got the getComplexType() methods to work without the unchecked cast

OhmV-IR (approved):
LGTM then

Seggan (changes requested):
I feel that the distinction between the map PDT and the map PDT provider is useless. Why not move all the logic into the map PDT and simply make new instances when needed?

Seggan (approved):
Hehe companion objects go brr

PR #19, opened 2025-01-25
Add test failure communication

Seggan:
The GH action/`runServer` task will now fail if any test failed. Also fixed tests deadlocking on error.

### DIFF
--- a/pylon-test/src/main/java/io/github/pylonmc/pylon/test/GenericTest.java
+++ b/pylon-test/src/main/java/io/github/pylonmc/pylon/test/GenericTest.java
@@ -7,5 +7,7 @@ public interface GenericTest {
         return new NamespacedKey(TestAddon.instance(), this.getClass().getSimpleName());
     }
     void run();
-    void cleanup();
+
+    default void cleanup() {
+    }
 }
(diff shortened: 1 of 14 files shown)


pylon-test/src/main/java/io/github/pylonmc/pylon/test/TestAddon.java
```diff
@@ -53,16 +56,22 @@ public void onEnable() {
...
                         logger.info("FAILED: %s".formatted(failedString));
-                    } else {
-                        logger.info("ALL TESTS PASSED");
+
+                        // Communicate back to the runServer task that the tests failed
+                        File file = new File("tests-failed");
+                        try {
+                            file.createNewFile();
+                        } catch (IOException e) {
+                            throw new UncheckedIOException(e);
```
LordIdra, line 73: Minor thing but if this is thrown, server won't be shut down

pylon-test/src/main/java/io/github/pylonmc/pylon/test/GenericTest.java
```diff
@@ -7,5 +7,7 @@ default NamespacedKey getKey() {
         return new NamespacedKey(TestAddon.instance(), this.getClass().getSimpleName());
     }
     void run();
-    void cleanup();
+
+    default void cleanup() {
```
LordIdra, line 11: could we also remove the cleanup() overrides from the current tests if we're making this default?

pylon-test/src/main/java/io/github/pylonmc/pylon/test/TestAddon.java
LordIdra: actually nvm probably best server hangs here so we know something is up, oitherwise tests just pass

LordIdra (approved):

Intybyte (approved):

PR #20, opened 2025-01-27
Add text hints for PylonBlock

OhmV-IR:
Text hints are initialized along with the PylonBlock class and stored as an accessible field in the PylonBlock class. These hints are activated when the player looks at the block and deactivated when the player looks away(listeners on TODO list with the other block listeners). blockHint styling and text are open vars so they can be changed after the hint has been initialized.


JustAHuman-xD:
Not a fan of this current implementation I won't lie, it's gonna need to be properly dynamic and based on *actual* targeted blocks, I am a fan of having an actual block hint class but it shouldn't at the lowest level just be a string, we also need to keep in mind how this system will likely need to be modular, so that super classes can have an effect without the extending class needing to do all the heavy lifting.

Lastly, I don't think implementing this should be optional, but I don't think it's a good idea to have it in the constructor especially in it's current form as *three* parameters. If my first pieces of advice are taken I imagine it will be removed from the constructor entirely which would solve this, but adding it as it's own point anyways.

JustAHuman-xD (changes requested):
(See the above)

Seggan (changes requested):
Yeah as Justin said this shouldn't be implemented as constructor parameters, I would suggest something closer to an `abstract val` or something

JustAHuman-xD:
> Yeah as Justin said this shouldn't be implemented as constructor parameters, I would suggest something closer to an `abstract val` or something

Even then, if it's gonna be block sensitive then that would still not work no?

LordIdra:
just my 2 cents but why not just have an overridable method on the block getWailaLine or something idk, and then implement the actual waila logic in an event handler? that just seems like by far the simplest and most flexible solution, subclasses of PylonBlock can then implement their own WAILA handlers if they want to give info on energy stored etc

LordIdra:
also there is no event telling us when the player stops/start looking at a block. we have to poll and check where they're looking continuously. that makes a cache (like in this PR) useless because it's not actually simplifying anything or saving performance

Seggan:
> > Yeah as Justin said this shouldn't be implemented as constructor parameters, I would suggest something closer to an `abstract val` or something
> 
> Even then, if it's gonna be block sensitive then that would still not work no?

Yeah, I originally proposed basically what Idra said

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/PylonBlock.kt
```diff
@@ -3,12 +3,20 @@ package io.github.pylonmc.pylon.core.block
...
+
+open class PylonBlock<S: PylonBlockSchema> private constructor(val schema: S, val block: Block, hintText: String, hintColor: BarColor, hintStyle: BarStyle) {
+    open val blockHint: PylonBlockHintText = PylonBlockHintText(hintText, hintColor, hintStyle)
 
-open class PylonBlock<S: PylonBlockSchema> private constructor(val schema: S, val block: Block) {
     constructor(reader: PylonDataReader, block: Block)
-            : this(getSchemaOfType<S>(reader.id), block)
+            : this(getSchemaOfType<S>(reader.id), block, block.type.toString(), BarColor.RED, BarStyle.SOLID)
+    constructor(schema: S, block: Block) : this(schema, block, block.type.toString(), BarColor.RED, BarStyle.SOLID)
+
```
Intybyte, line 19: you created PylonBlockHintText just to not use it lol, just wrap these 3 fields into a PylonBlockHintText

OhmV-IR: done

Intybyte:
> just my 2 cents but why not just have an overridable method on the block getWailaLine or something idk, and then implement the actual waila logic in an event handler? that just seems like by far the simplest and most flexible solution, subclasses of PylonBlock can then implement their own WAILA handlers if they want to give info on energy stored etc

I would prefer this approach too, just read the comments now I just jumped to review on this one idk why, tough it might not be as simple as you need to pass extra data so maybe make it return like BossBar straight up, I don't think a wrapper like Ohm's PylonBlockHintText is really needed in we take this approach.

It would still need to be handled based on the location, so maybe blockDisplayInfoAt(Player, Location) would work

JustAHuman-xD:
Why location and not Block? Also I do think we should have our own wrapper around the system for modularity & resource pack integration down the line.

JustAHuman-xD:
> Why location and not Block? Also I do think we should have our own wrapper around the system for modularity & resource pack integration down the line. 

Though that could just be changed once it's needed rather than now, that wouldn't be too bad.

OhmV-IR:
> > just my 2 cents but why not just have an overridable method on the block getWailaLine or something idk, and then implement the actual waila logic in an event handler? that just seems like by far the simplest and most flexible solution, subclasses of PylonBlock can then implement their own WAILA handlers if they want to give info on energy stored etc
> 
> I would prefer this approach too, just read the comments now I just jumped to review on this one idk why, tough it might not be as simple as you need to pass extra data so maybe make it return like BossBar straight up, I don't think a wrapper like Ohm's PylonBlockHintText is really needed if we take this approach
> 
> It would still need to be handled based on the location, so maybe blockDisplayInfoAt(Player, Location) would work

I've done an implementation of this by creating the BlockHint interface, got rid of PylonBlockHintText class and just integrated the bossbar into the PylonBlock class itself

Seggan:
If you rebase on [item-and-block-consistency](https://github.com/pylonmc/pylon-core/tree/item-and-block-consistency) you can make `getHint` an `abstract val` instead of constructor parameter

JustAHuman-xD:
> That still doesn't solve most of the other issues brought up though?

Actually thats wrong, but it doesn't solve the glaring problem of not having the context of the block being looked at

Intybyte:
> > That still doesn't solve most of the other issues brought up though?
> 
> Actually thats wrong, but it doesn't solve the glaring problem of not having the context of the block being looked at

Can you tell us more about what context might be useful?

Intybyte:
Also passing Block/Location and Player is necessary, considering that we might want to display energy of a machine and more location specific data, such as a solar panel working (if exposed to sky) etc

JustAHuman-xD:
> Also passing Block/Location and Player is necessary, considering that we might want to display energy of a machine and more location specific data, such as a solar panel working (if exposed to sky) etc

Exactly, also translations, if we don't know the player we don't know the language, etc

OhmV-IR:
Closing, going to leave this for later once more systems are built up

PR #22, opened 2025-01-28
Switch to coroutines

Seggan:
The above. Also switched out `synchronized (this)` for `ReadWriteLock`


LordIdra (approved):
can't find any issues, thanks for this

PR #23, opened 2025-01-29
Consistency-ise block and item classes

LordIdra:
I noticed some problems with block and item [schemas] while testing. PylonBlock and PylonItem are now abstract and have corresponding simple implementations. PylonItem is now parameterized by schema and also now stores a schema when loaded, this was a big oversight in my initial PR

### DIFF
--- a/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/item/PylonItem.kt
+++ b/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/item/PylonItem.kt
@@ -8,9 +8,9 @@ import org.bukkit.inventory.ItemStack
 
 abstract class PylonItem<out S: PylonItemSchema>(
     val schema: S,
-    stack: ItemStack
-) : ItemStack(stack) {
-    val id = persistentDataContainer.get(idKey, PylonSerializers.NAMESPACED_KEY)!!
+    val stack: ItemStack
+) {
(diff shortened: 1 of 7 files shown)


Seggan (approved):

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/item/PylonItem.kt
```diff
@@ -6,11 +6,14 @@ import io.github.pylonmc.pylon.core.registry.PylonRegistry
 import org.bukkit.NamespacedKey
 import org.bukkit.inventory.ItemStack
 
-open class PylonItem(stack: ItemStack) : ItemStack(stack) {
+abstract class PylonItem<out S: PylonItemSchema>(
+    val schema: S,
+    stack: ItemStack
+) : ItemStack(stack) {
```
Intybyte, line 12: why is this extending ItemStack in the first place...

Intybyte (changes requested):

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/item/PylonItem.kt
LordIdra: why wouldn't it..?

Seggan: because you dont extend itemstack. this is literally the issue that slimefun had

PR #26, opened 2025-02-05
Add recipe framework

Seggan:
Still gotta add the default recipe types for stuff like vanilla crafting and furnace


pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/recipe/MobDropRecipe.kt
```diff
@@ -0,0 +1,35 @@
...
+import org.bukkit.entity.EntityType
+import org.bukkit.entity.LivingEntity
+import org.bukkit.inventory.ItemStack
+import java.util.concurrent.ThreadLocalRandom
+
+interface MobDropRecipe : Keyed {
+
+    fun getResult(entity: LivingEntity, playerKill: Boolean): ItemStack?
+
+    class Simple @JvmOverloads constructor(
```
LordIdra, line 14: Maybe this would be better as an externa class SimpleMobDropRecipe just to keep it consistent with SimplePylonBlock / SimplePylonItem?

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/recipe/MobDropRecipe.kt
```diff
@@ -0,0 +1,35 @@
+package io.github.pylonmc.pylon.core.recipe
+
+import org.bukkit.Keyed
+import org.bukkit.NamespacedKey
+import org.bukkit.entity.EntityType
+import org.bukkit.entity.LivingEntity
+import org.bukkit.inventory.ItemStack
+import java.util.concurrent.ThreadLocalRandom
+
+interface MobDropRecipe : Keyed {
```
LordIdra, line 10: What's the reason for this being an interface?

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/recipe/RecipeType.kt
```diff
@@ -0,0 +1,16 @@
+package io.github.pylonmc.pylon.core.recipe
+
+import io.github.pylonmc.pylon.core.registry.PylonRegistry
+import org.bukkit.Keyed
+import org.bukkit.NamespacedKey
+
+interface RecipeType<T : Keyed> : Keyed, Iterable<T> {
```
LordIdra, line 7: I assume this is an interface because of the vanilla recipes. Would it be worthwhile to create a SimpleRecipeType class that just has a list inside - like MobDropRecipeType does - to reduce boilerplate?

LordIdra: Or actually maybe have RecipeType be an object, then VanillaRecipeType extends that and overrides the methods? Simpler/clearer IMO

LordIdra, line 7: Could we also maybe get a getRecipes(): List\<T\> function or something similar? The iterator is nice in some situations but it can be a pain when for example you want to

LordIdra (changes requested):
lgtm just commented on a couple things

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/recipe/MobDropRecipe.kt
Seggan: So that people can implement it with their own constraints

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/recipe/RecipeType.kt
Seggan: In general when I'm looking for something that people are intended to extend, I reach for an interface first

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/recipe/MobDropRecipe.kt
Intybyte, line 10: I think it is better to keep this separate from the recipe system, maybe a loot table registry/handling, we aren't forced to follow the weird way slimefun handles RecipeTypes where we need to have them merged

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/recipe/RecipeTypes.kt
```diff
@@ -0,0 +1,57 @@
...
+
+        override fun removeRecipe(recipe: NamespacedKey) {
+            Bukkit.removeRecipe(recipe)
+            recipes.removeIf { it.key == recipe }
+        }
+
+        override fun iterator(): Iterator<T> = recipes.iterator()
+
+        override fun getKey(): NamespacedKey = key
+    }.also { it.register() }
```
Intybyte, line 56: It would be beneficial to extract this into its own class for other people to use

Intybyte (changes requested):

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/recipe/RecipeTypes.kt
Seggan: There's like no other vaniall recipe types that can be wrapped like this, why would I do that?

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/recipe/MobDropRecipe.kt
LordIdra: at that point wouldn't you just create a new recipe type and implement that? I would have never thought to implement the existing recipe type if I was an addon developer

Seggan: If I were to make something that only dropped on a full moon, I would definetly not use a whole new recipe type, and registering my own listeners, and whatnot

LordIdra: Idk it just seems a bit weird to have an interface for a recipe when your own recipe class with a listener will do just fine. Smells like over-abstraction. But I won't block PR on it

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/recipe/RecipeTypes.kt
Intybyte: oh i guess it is fine then

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/recipe/MobDropRecipe.kt
Intybyte: Vanilla Minecraft treats mob drops as separate from crafting
If you register them as recipes, it may confuse developers using your framework (i am still kinda confused by how sf does it)
Mob dropping is often used with event unlike recipes which require the player to assemble stuff in a grid, smth like EntityDeathEvent, it makes little sense to have them as one thing, if anything i believe as it uses events to be more fitting to be handled by han interface just like the other events

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/recipe/MobDrop.kt
```diff
@@ -0,0 +1,10 @@
+package io.github.pylonmc.pylon.core.recipe
+
+import org.bukkit.Keyed
+import org.bukkit.entity.LivingEntity
+import org.bukkit.inventory.ItemStack
+
+interface MobDrop : Keyed {
+
+    fun getResult(entity: LivingEntity, playerKill: Boolean): ItemStack?
+}
```
Intybyte, line 10: Just pass the whole event, no reason to exposed a limited set of info

Intybyte (changes requested):

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/recipe/MobDrop.kt
Intybyte: also at this point mobdrop should have their own package

LordIdra (approved):
sorry for the wait, completely forgot to review this yesterday lol, lgtm

PR #29, opened 2025-02-28
Add block infrastructure

Seggan:
I was peacefully working on some pylon-base blocks, so imagine my surprise when I realized that we can't even place blocks yet. This does that, and also adds a few utils I needed for pylon-base.

Also we need to annotate all the `internal` goodness with `@JvmSynthetic`, that will be in a separate PR.

PR #30, opened 2025-02-28
Add `@JvmSynthetic` everywhere

Seggan:
For reasons see https://discord.com/channels/1329177304857055273/1329177972631928915/1345130973809082381

Merge after #29

### DIFF
--- a/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/PylonCore.kt
+++ b/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/PylonCore.kt
@@ -1,6 +1,9 @@
+@file:JvmSynthetic // Hide `PylonCoreKt.getPluginInstance` from Java
+
 package io.github.pylonmc.pylon.core
 
 import co.aikar.commands.PaperCommandManager
+import io.github.pylonmc.pylon.core.addon.PylonAddonListener
 import io.github.pylonmc.pylon.core.block.BlockListener
 import io.github.pylonmc.pylon.core.block.TickManager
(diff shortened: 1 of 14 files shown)


Intybyte (approved):
9mcbkc

OhmV-IR:
I’d like to have some kind of tool-based way to check this and add the tag for anything that isn’t explicitly marked public / open in kotlin, either as a GH actions check or a Python script. The ugly approach will do for the time being though, and that could be added to the chore list.

OhmV-IR (approved):
LGTM CW assuming you’ve done a compile check on both core **and** base to make sure you haven’t accidentally hidden anything important.

Seggan:
> LGTM CW assuming you’ve done a compile check on both core **and** base to make sure you haven’t accidentally hidden anything important.

If base breaks it's because it was supposed to be hidden anyway

Seggan:
> I’d like to have some kind of tool-based way to check this and add the tag for anything that isn’t explicitly marked public / open in kotlin, either as a GH actions check or a Python script. The ugly approach will do for the time being though, and that could be added to the chore list.

This is only for `internal`, as Kotlin's implementation of internal is stupid

OhmV-IR:
> > I’d like to have some kind of tool-based way to check this and add the tag for anything that isn’t explicitly marked public / open in kotlin, either as a GH actions check or a Python script. The ugly approach will do for the time being though, and that could be added to the chore list.
> 
> This is only for `internal`, as Kotlin's implementation of internal is stupid

still a chore I suppose though

LordIdra (approved):

PR #43, opened 2025-03-25
Add config system

LordIdra:
- Adds a Config class for accessing yml configs
- Adds a ConfigSection class which does what you think it does
- Adds PylonConfig which stores config values for pylon

### DIFF
--- a/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/config/ConfigSection.kt
+++ b/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/config/ConfigSection.kt
@@ -5,7 +5,7 @@ import org.bukkit.configuration.ConfigurationSection
 open class ConfigSection(val internalSection: ConfigurationSection) {
 
     class KeyNotFoundException(path: String?, key: String)
-        : Exception(if (path != null) { "Config key not found: $path.$key" } else { "Config key not found: $key"})
+        : Exception(if (path != null) "Config key not found: $path.$key" else "Config key not found: $key")
 
     fun getSections(): Set<ConfigSection> {
         val configSections: MutableSet<ConfigSection> = mutableSetOf()
@@ -26,16 +26,16 @@ open class ConfigSection(val internalSection: ConfigurationSection) {
     fun getSection(key: String): ConfigSection
         = getSectionMaybe(key) ?: throw KeyNotFoundException(internalSection.currentPath, key)
 
-    inline fun <reified T> getMaybe(key: String): T? {
+    inline fun <reified T> get(key: String): T? {
         val value = internalSection.get(key) ?: return null
         return value as T
     }
 
-    inline fun <reified T> get(key: String): T
-        = getMaybe(key) ?: throw KeyNotFoundException(internalSection.currentPath, key)
+    inline fun <reified T> getOrThrow(key: String): T
+        = get(key) ?: throw KeyNotFoundException(internalSection.currentPath, key)
 
     inline fun <reified T> get(key: String, default: T): T
-        = getMaybe(key) ?: default
+        = get(key) ?: default
 
     fun <T> set(key: String, value: T) {
         internalSection.set(key, value)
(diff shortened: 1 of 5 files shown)


pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/config/ConfigSection.kt
```diff
@@ -0,0 +1,43 @@
...
+
+    inline fun <reified T> getMaybe(key: String): T? {
+        val value = internalSection.get(key) ?: return null
+        return value as T
+    }
+
+    inline fun <reified T> get(key: String): T
+        = getMaybe(key) ?: throw KeyNotFoundException(internalSection.currentPath, key)
+
+    inline fun <reified T> get(key: String, default: T): T
```
Seggan, line 37: Instead of `getMaybe` and `get`, I'd suggest `get` and `getOrThrow`

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/config/ConfigSection.kt
```diff
@@ -0,0 +1,43 @@
+package io.github.pylonmc.pylon.core.config
+
+import org.bukkit.configuration.ConfigurationSection
+
+open class ConfigSection(val internalSection: ConfigurationSection) {
```
Seggan, line 5: Why is this `open`?

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/config/ConfigSection.kt
```diff
@@ -0,0 +1,43 @@
...
+
+    class KeyNotFoundException(path: String?, key: String)
+        : Exception(if (path != null) { "Config key not found: $path.$key" } else { "Config key not found: $key"})
```
Seggan, line 8:
```suggestion
        : Exception(if (path != null) "Config key not found: $path.$key" else "Config key not found: $key")
```

Seggan (changes requested):
The lore of the ancients is accurate: "Verily, the briefer a pull request be, the more rebuke it receiveth."

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/config/ConfigSection.kt
LordIdra: Allows people to add their own stuff, for example a getMaterial method. I had to do this with aircraft

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/config/ConfigSection.kt
```diff
@@ -0,0 +1,43 @@
...
+
+    fun getSectionNames(): Set<String>
+        = internalSection.getKeys(false)
+
+    fun getSectionMaybe(key: String): ConfigSection? {
+        val newConfig = internalSection.getConfigurationSection(key) ?: return null
+        return ConfigSection(newConfig)
+    }
+
+    fun getSection(key: String): ConfigSection
```
Seggan, line 26: Same here, didn't realize it was here

Seggan (changes requested):

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/config/ConfigSection.kt
LordIdra: oh oops

Seggan (dismissed):

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/config/PylonConfig.kt
```diff
@@ -0,0 +1,14 @@
+package io.github.pylonmc.pylon.core.config
+
+import io.github.pylonmc.pylon.core.pluginInstance
+
+object PylonConfig {
+
+    private val config = Config(pluginInstance, "config.yml")
```
OhmV-IR, line 7: config.yml is too generic of a name, it's likely to cause conflicts with other things. I would recommend pylonConfig.yml

OhmV-IR (changes requested):
one thing

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/config/PylonConfig.kt
Seggan: That's the default config name for any plugin. You can't mess it up since it's in a subdir

OhmV-IR (approved):
LGTM CW, have not tested

PR #44, opened 2025-03-25
detect pylon item entering / leaving inventory

OhmV-IR:
Will enable passive buffs such as a healing talisman to give you health while it is in your inventory. Using a runnable that runs every 15 ticks instead of events because it is actually more performant than handling all the events that can cause an item pickup.


LordIdra (changes requested):
I don't think this is the best way to implement inventory-bound items:
1) What we're more likely to want is some kind of ticker, because that's far more broad. Think for example a talisman that only activates when you fall into lava, or a jetpack which deactivates when it runs out of fuel, or a machine which automatically feeds you every so often. This item is a bit of a special case
2) What you're trying to do is really hard to get right. How does it work with multiple items throughout the inventory? What about multiple items in one stack? What if the player leaves/rejoins? What if the server restarts? What if the server hard crashes? How can we tell it's actually the same item and not a different one that was swapped out and maybe has different data on it? 
3) As mentioned - this is a bit of a special case, so I think it's best to either close this PR, or change this PR to add an inventory item ticker interface instead. Then, the talisman implementation can use a different method - for example, you could set a special 'health_boosted' key on the player's PDC container, and then use it to check when a player has added/removed a talisman. This would probably be simpler and a lot less error-prone

OhmV-IR:
Looking to do the ticker interface, will request rereview when I have something working, will likely heavily inherit from the block ticker where possibel

OhmV-IR:
alright have the item ticker iface implemented and the health talisman base pr has been updated to use the new setup

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/item/ItemTickerRunnable.kt
```diff
@@ -0,0 +1,17 @@
...
+import org.bukkit.scheduler.BukkitRunnable
+
+class ItemTickerRunnable : BukkitRunnable() {
+    override fun run() {
+        for (itemEntry in PylonRegistry.ITEMS){
+            val pylonItem = PylonItem.fromStack(itemEntry.itemStack)
+            if(pylonItem is TickingItem){
+                pylonItem.tick();
+            }
+        }
```
Seggan, line 14: I'm a bit confused. Isn't this supposed to run for inventories?

Seggan (changes requested):

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/item/ItemTickerRunnable.kt
Seggan: But the ticking logic is called for a brand new item, what item is supposed to tick if it's not in an inventory?

Seggan: I cannot think of a use case where the item wouldnt need to be in an inventory. Additionally, the performance penalties would be actually worse as each item will iterate over the inventory by itself. I think this is what Idra was saying.

OhmV-IR: then how is the talisman going to work? the item can't detect when it gets removed from the inventory, and this is just going back to the system I had before in commit 297cc74ce3407cf3b70f88f649cc28594a21f5df

Seggan: You would possibly use a delayed task to check whether the talisman is not in the inv, then remove effects if it's not

OhmV-IR: that then eliminates the need for a core PR because the ticker can be contained entirely within base

PR #48, opened 2025-03-30
Fix recipe matching

OhmV-IR:
As shown on discord, fixes the PylonItem -> VanillaItem crafting protection by comparing keys instead of Recipe objects, still don't know why the matching wasn't working properly on it though.

### DIFF
--- a/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/recipe/RecipeTypes.kt
+++ b/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/recipe/RecipeTypes.kt
@@ -55,8 +55,10 @@ private object CraftingRecipeType : VanillaRecipe<CraftingRecipe>("crafting") {
     @EventHandler(priority = EventPriority.LOWEST)
     private fun onPreCraft(e: PrepareItemCraftEvent) {
         val recipe = e.recipe ?: return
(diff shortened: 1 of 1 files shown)


pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/recipe/RecipeTypes.kt
```diff
@@ -83,8 +85,10 @@ private object SmithingRecipeType : VanillaRecipe<SmithingRecipe>("smithing") {
     @EventHandler(priority = EventPriority.LOWEST)
     private fun onSmith(e: PrepareSmithingEvent) {
         val inv = e.inventory
+        val recipe = inv.recipe
+        if(recipe == null || recipe !is Keyed) return
```
Seggan, line 89:
```suggestion
        if(recipe !is Keyed) return
```

Not blocking but the `is` checks nullable as well

Seggan (approved):

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/recipe/RecipeTypes.kt
OhmV-IR: did for both

Seggan (approved):

PR #54, opened 2025-04-01
Add various block events

OhmV-IR:
Ads a whole ton of interfaces for blocks to implement, works like PylonItems. Currently WIP. I'm sorry for whoever has to review about 40 of the same function over and over again

### DIFF
--- a/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/base/Beacon.kt
+++ b/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/base/Beacon.kt
@@ -1,0 +1,13 @@
+package io.github.pylonmc.pylon.core.block.base
+
+import com.destroystokyo.paper.event.block.BeaconEffectEvent
+import io.papermc.paper.event.block.BeaconActivatedEvent
+import io.papermc.paper.event.block.BeaconDeactivatedEvent
+import io.papermc.paper.event.player.PlayerChangeBeaconEffectEvent
+
+interface Beacon {
+    fun onActivated(event: BeaconActivatedEvent)
+    fun onDeactivated(event: BeaconDeactivatedEvent)
+    fun onEffectChange(event: PlayerChangeBeaconEffectEvent)
+    fun onEffectApply(event: BeaconEffectEvent)
+}
--- a/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/BlockListener.kt
+++ b/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/BlockListener.kt
@@ -8,6 +8,7 @@ import io.github.pylonmc.pylon.core.util.position.position
 import io.papermc.paper.event.block.BeaconActivatedEvent
 import io.papermc.paper.event.block.BeaconDeactivatedEvent
 import io.papermc.paper.event.block.CompostItemEvent
+import io.papermc.paper.event.block.PlayerShearBlockEvent
 import io.papermc.paper.event.block.TargetHitEvent
 import io.papermc.paper.event.entity.EntityCompostItemEvent
 import io.papermc.paper.event.player.PlayerChangeBeaconEffectEvent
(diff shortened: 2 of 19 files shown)


pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/base/Beacon.kt
```diff
@@ -0,0 +1,13 @@
+package io.github.pylonmc.pylon.core.block.base
+
+import com.destroystokyo.paper.event.block.BeaconEffectEvent
+import io.papermc.paper.event.block.BeaconActivatedEvent
+import io.papermc.paper.event.block.BeaconDeactivatedEvent
+import io.papermc.paper.event.player.PlayerChangeBeaconEffectEvent
+
+interface Beacon {
+    fun onActivated(event: BeaconActivatedEvent)
```
Seggan, line 9:
```suggestion
    fun onActivated(event: BeaconActivatedEvent) {}
```
These should all be default empty so that we won't have to do empty impls

Seggan (changes requested):

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/base/Beacon.kt
OhmV-IR: at least find and replace exists...

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/item/PylonItemListener.kt
```diff
@@ -62,6 +63,10 @@ internal object PylonItemListener : Listener {
         if (pylonItem is Interactor) {
             pylonItem.onUsedToRightClick(event)
         }
+        val pylonBlock = BlockStorage.get(event.clickedBlock ?: return)
+        if(pylonBlock is InteractableBlock){
+            pylonBlock.onInteract(event)
+        }
     }
```
LordIdra, line 70: This should be in BlockListener, since PylonItemListener is specifically to do with Pylon items. Having it here makes it look like this is related to pylon items

LordIdra (changes requested):
I think it would be worth changing the interfaces to be named eg PylonBeacon instead of Beacon, to avoid confusion and potential conflicts with Bukkit

Seggan (changes requested):
Can you add default impls for the existing interfaces too while ure at it

Seggan:
No that's not what I meant. I mean make all the interface methods default empty cause we often have to override and leave them empty

OhmV-IR:
> No that's not what I meant. I mean make all the interface methods default empty cause we often have to override and leave them empty

The only one I did it for is onBreak in BreakHandler, all the others either have them already or shouldn't have them(eg. tick in Ticking)

Seggan (approved):

LordIdra:
> I think it would be worth changing the interfaces to be named eg PylonBeacon instead of Beacon, to avoid confusion and potential conflicts with Bukkit

Do we think it's worth doing this for items as well to keep things consistent?

OhmV-IR:
> > I think it would be worth changing the interfaces to be named eg PylonBeacon instead of Beacon, to avoid confusion and potential conflicts with Bukkit
> 
> Do we think it's worth doing this for items as well (and entities in the entities PR) to keep things consistent?

I think it's worth doing, and since this PR is already modifying the item ifaces to include the default method bodies, it might as well change the names as well

OhmV-IR:
> > I think it would be worth changing the interfaces to be named eg PylonBeacon instead of Beacon, to avoid confusion and potential conflicts with Bukkit
> 
> Do we think it's worth doing this for items as well (and entities in the entities PR) to keep things consistent?

I think we should do it for items too, but it would be breaking for all the current base PRs, so I'll make an issue for it and mark it as a chore

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/item/PylonItemListener.kt
LordIdra: (I'm happy with this PR once this is fixed)

LordIdra:
Could we get the existing block interfaces renamed to PylonXYZ? Also the interfaces with just one method should not have a default implementation

LordIdra (approved):

PR #55, opened 2025-04-01
Test improvements

LordIdra:
Big test refactor

- Replaces this common pattern with `runSync` and `runAsync` functions
```java
CompletableFuture<Void> future = new CompleteableFuture();
Bukkit.getScheduler().runTask(() -> {
    doSomething();
    future.complete();
});
future.join();
```
- Lots of tiny formatting improvements and warning fixes
- Log successes as well as failure
- Extract out blocks from their corresponding tests to allow reuse across tests
- Fixes issues where BlockStorage checks would load chunks
- Attempting to get Pylon blocks in unloaded chunks now throws an IllegalStateException
- Adds async helper methods (runSync, runAsync, loadChunk, unloadChunk)
- Massively simplified BlockStorage tests by removing all the listener logic and using the new helper methods. (This has also resulted in a considerable speedup)
- Measures the time each test takes


Seggan:
I'm sorry for what I'm about to do

OhmV-IR:
> I'm sorry for what I'm about to do

NO SEGGAN YOU CANT GIVE IN

Seggan (approved):
Uhh since apparently I'm dead, consider this approved

PR #56, opened 2025-04-02
Implement item settings

Seggan:
Companion Base PR Coming Soon™

### DIFF
--- a/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/addon/PylonAddon.kt
+++ b/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/addon/PylonAddon.kt
@@ -47,6 +47,8 @@ interface PylonAddon : Keyed {
         val resource = this.javaPlugin.getResource(path)
         if (resource != null) {
             val newConfig = resource.reader().use(YamlConfiguration::loadConfiguration)
+            config.internalConfig.setDefaults(newConfig)
+            config.internalConfig.options().copyDefaults(true)
             config.merge(ConfigSection(newConfig))
             config.save()
         }
--- a/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/addon/PylonAddon.kt
+++ b/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/addon/PylonAddon.kt
@@ -1,8 +1,12 @@
 package io.github.pylonmc.pylon.core.addon
 
+import io.github.pylonmc.pylon.core.config.Config
+import io.github.pylonmc.pylon.core.config.ConfigSection
+import io.github.pylonmc.pylon.core.pluginInstance
 import io.github.pylonmc.pylon.core.registry.PylonRegistry
 import org.bukkit.Keyed
 import org.bukkit.NamespacedKey
+import org.bukkit.configuration.file.YamlConfiguration
 import org.bukkit.plugin.java.JavaPlugin
 
 interface PylonAddon : Keyed {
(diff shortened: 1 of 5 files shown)


pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/config/ConfigSection.kt
```diff
@@ -1,11 +1,14 @@
 package io.github.pylonmc.pylon.core.config
 
+import com.google.common.base.CaseFormat
 import org.bukkit.configuration.ConfigurationSection
+import kotlin.properties.ReadWriteProperty
+import kotlin.reflect.KProperty
 
-open class ConfigSection(val internalSection: ConfigurationSection) {
```
LordIdra, line 5: I think this is something we should expose in case someone wants to do something that requires a ConfigurationSection - I've had this issue before when I implemented a similar system, which is why I didn't make it private

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/item/PylonItemSchema.kt
```diff
@@ -36,7 +37,27 @@ open class PylonItemSchema(
...
+            val newConfig = resource.reader().use { ConfigSection(YamlConfiguration.loadConfiguration(it)) }
+            settings.merge(newConfig)
+            settings.save()
+        }
+    }
+
     fun register() = apply {
+        ItemStackBuilder(template) // Modifies the template directly
+            .editMeta { meta -> meta.persistentDataContainer.set(idKey, PylonSerializers.NAMESPACED_KEY, key) }
+            .lore(LoreBuilder().addon(addon))
```
LordIdra, line 60: What's the reason for this change?

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/item/PylonItemSchema.kt
```diff
@@ -36,7 +37,27 @@ open class PylonItemSchema(
...
+    val settings: Config
+
+    init {
+        val settingsFile = getItemFile(key)
+        if (!settingsFile.exists()) {
+            settingsFile.parent.createDirectories()
+            settingsFile.createFile()
+        }
+        settings = Config(settingsFile.toFile())
+        val resource = addon.javaPlugin.getResource("settings/item/${keyPath(key)}")
```
LordIdra, line 49: The combination of settingsFile, itemsDir, keyPath, and a hardcoded string here is really confusing. Couldn't `itemsDir` and `keyPath` just be turned into a single `itemPath(key: NamespacedKey): String`? I don't think we need the whole `itemsDir.createDirectories()` thing given getItemFile already does `path.parent.createDirectories()` - or if it's still needed, that should be merged together IMO

LordIdra (changes requested):

LordIdra:
Also - will block settings be a separate PR then?

Seggan:
> Also - will block settings be a separate PR then?


Yes

OhmV-IR (changes requested):
I think file paths are broken on windows again, and the fix will likely have to be implemented here so hold on merging this

Seggan:
> I think file paths are broken on windows again, and the fix will likely have to be implemented here so hold on merging this

Already replied on discord

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/addon/PylonAddon.kt
```diff
@@ -22,4 +26,32 @@ interface PylonAddon : Keyed {
      * The display name used, for example, at the bottom of items to show which addon an item is from
      */
     fun displayName(): String
+
+    /**
+     * Merges config from resources to the Pylon config directory.
```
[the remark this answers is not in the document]
LordIdra: we're not really following javadoc convention anywhere, I don't think we should use it here, we can worry about this once we start proper docs

PR #59, opened 2025-04-04
Fix occasional BlockStorageChunkReloadTest failure

LordIdra:
I believe the issue with this test was that the chunks were adjacent, so there was some weird behaviour that meant some of them would get loaded by other adjacent chunks while chunk generation was happening. This changes the chunks to be random. Unfortunately this does mean we can only test 20 chunks rather than 400 (due to performance issues) but this isn't really a big deal


OhmV-IR (approved):
LGTM have tested on my machine, ran 5 tests w this pr and all passed

PR #62, opened 2025-04-08
Add pylondebug command

OhmV-IR:
Gives you the pylon debug waxed weathered copper cut stairs:tm: to fulfill all your debugging needs.

Command is /py debug or /pylon debug, takes no args and gives the aforementioned debug item to the player who executes the command


OhmV-IR:
closing because the branch accidentally was made off of the entities one instead of master

PR #65, opened 2025-04-09
Allow entity holder block to hold multiple entities

LordIdra:


Seggan (dismissed):

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/base/EntityHolderBlock.kt
```diff
@@ -5,49 +5,72 @@ import io.github.pylonmc.pylon.core.entity.PylonEntity
...
+ * Note that the Pylon entities may not be loaded when the block is loaded.
  */
-interface EntityHolderBlock<out E : PylonEntity<*, *>> : PylonBreakHandler {
+interface EntityHolderBlock : PylonBreakHandler {
 
     /**
      * Must be set in your create constructor when you spawn in the entity.
      */
-    var entityUuid: UUID
+    val heldEntities: Map<String, UUID>
```
OhmV-IR, line 24: instead of using a string name, what about using a NamespacedKey? also re my comment on the entities PR about making this a getter so that you are forced to override it by the compiler

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/entity/display/transform/TransformBuilder.kt
```diff
@@ -118,6 +118,6 @@ class TransformBuilder(val components: ArrayDeque<TransformComponent>) {
     }
 
     companion object {
-        private val BLOCK_DISPLAY_ADJUSTMENT: Vector3f = Vector3f(-0.5F)
+        private val BLOCK_DISPLAY_ADJUSTMENT: Vector3f = Vector3f(-0.5F, -0.5F, -0.5F)
```
OhmV-IR, line 121: same comment is on entities PR, I think plugins would find having this value useful, so
```suggestion
        @JvmField
        val BLOCK_DISPLAY_ADJUSTMENT: Vector3f = Vector3f(-0.5F, -0.5F, -0.5F)
```

OhmV-IR (changes requested):
Some stuff to consider

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/base/EntityHolderBlock.kt
LordIdra: See reply on other PR re making it a getter. There's no reason to use NamespacedKey here because this is local to a single block implementation - other addons will not be inserting stuff here, so there's really no need for a namespace. Strings are just easier to work with.

PR #66, opened 2025-04-13
Hard fail when schema not registered

LordIdra:
Tiny PR, just makes blocks and entities throw an exception if you attempt to create them using a schema that hasn't been registered. This has caught me out on several occasions.


LordIdra:
ugh this breaks phantom blocks

OhmV-IR (approved):
LGTM, tested by running base, then commenting the registering of some entities/blocks and running again, can't really test PhantomBlock easily so I'm not going to worry about it

PR #67, opened 2025-04-13
Add lombok to pylon-test

LordIdra:
Does what it says on the tin


OhmV-IR (approved):
LGTM, haven't tested ig but if the server compiles and starts enough for tests to run it'll be fine

PR #70, opened 2025-04-15
Add IMMEDIATE_FACES and IMMEDIATE_FACES_WITH_DIAGONALs to Network.kt

OhmV-IR:


Seggan (changes requested):
These should be put into their own files

LordIdra (approved):

PR #73, opened 2025-04-17
Add WAILA

Seggan:
What am I looking at, you ask? Some of the best code ever written.


pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/command/PylonCommand.kt
```diff
@@ -58,6 +59,18 @@ object PylonCommand : BaseCommand() {
         BlockStorage.placeBlock(location, pylonBlock)
     }
 
+    @Subcommand("waila")
+    @Description("Set WAILA for this player")
```
LordIdra, line 63: Maybe 'Toggle your WAILA bossbar' (player-centric because it'll be used by players not admins, and hints what WAILA is, because many people might not know)

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/command/PylonCommand.kt
```diff
@@ -58,6 +59,18 @@ object PylonCommand : BaseCommand() {
...
+    @CommandPermission("pylon.command.waila")
+    fun waila(player: Player, enabled: Boolean) {
+        Waila.setWailaEnabled(player, enabled)
+        if (enabled) {
+            Waila.addPlayer(player)
+        } else {
+            Waila.removePlayer(player)
```
LordIdra, line 70: I think it may be better to encapsulate this logic in setWailaEnabled itself so it always takes instant effect

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/i18n/packet/PacketManager.kt
```diff
@@ -4,6 +4,7 @@ import io.github.pylonmc.pylon.core.i18n.PlayerTranslationHandler
 import org.bukkit.entity.Player
 import org.jetbrains.annotations.ApiStatus
 
+@ApiStatus.Internal
```
LordIdra, line 7: Wait why not `internal interface`, am I being thick

pylon-core/src/main/resources/config.yml
```diff
@@ -2,4 +2,7 @@
 tick-delay: 1
 
 # The number of errors a ticking block can throw before it is disabled
-allowed-block-errors: 5
\ No newline at end of file
+allowed-block-errors: 5
+
+# The interval, in ticks, between WALIA checks
+walia-interval: 5
```
LordIdra, line 8: worth setting this to 1 by default? I can't imagine WAILA is going to be heavy (and we can change the default later if we find out otherwise) and remember there's often 200ms or more of latency so doing this quickly is a big plus

LordIdra (changes requested):
Looks solid, just a few small things

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/i18n/packet/PacketManager.kt
Seggan: It won't be accessible for extension to `PacketManagerImpl`

LordIdra (approved):

JustAHuman-xD (approved):

PR #75, opened 2025-04-17
Obliterate simple item schema

LordIdra:
No longer needed after recipe refactor


Seggan (approved):

Seggan:
Just one comment: you sure we shouldn't keep this around for other addons?

LordIdra:
Don't think there's any reason to keep it in core. They can add it if they really want it

PR #96, opened 2025-05-01
Support cooldowns

Seggan:


LordIdra (commented):
Could you detail more the motivation for this? Why not use player.setCooldown like we do in hammer, and why is this inherited by BlockPlacer, BlockInteractor, EntityInteractor, and Interactor?

Seggan:
Basically, I'm currently working on the loupe, which also requires a cooldown. However, for some reason, `PlayerINteractEvent` doesn't respect it, so we have to manually do that.

LordIdra:
Makes sense. Why is it extended by BlockPlacer, BlockInteractor, EntityInteractor, and Interactor though?

Seggan:
BlockPlacer, BlockInteractor, and Interactor all use the same event, so I thought it'd be nice to have one central check in the event handler instead of disparate properties in every handler interface. EntityInteractor is for completeness/because I can

OhmV-IR (approved):
fixing mojang's problems yet again

LordIdra (approved):

PR #99, opened 2025-05-02
Fix wrapping

Seggan:
Yes


pylon-core/src/test/kotlin/io/github/pylonmc/pylon/core/i18n/wrapping/WrapTest.kt
```diff
@@ -0,0 +1,56 @@
+package io.github.pylonmc.pylon.core.i18n.wrapping
```
LordIdra, line 1: not super happy with WrapTest since it means we now have two different ways to test stuff. I assume this was necessary to access internals?

Seggan: No, it's cause spinning up a new server every time and waiting for all tests was too tedious, the unit tests finish practically instantly

LordIdra: Makes sense. do you think we could move them to PylonTest now that you're no longer iterating with them? it just makes me very uneasy if there are two different ways to test things. if we do go this route we should add CI for this at least

Seggan: Yeah I'll move it back, but ill keep the test deps and stuff for any future situations like this

LordIdra (approved):

PR #101, opened 2025-05-12
Fluids

LordIdra:
Landmark PR to add a fluid system. Corresponding base PR is not ready yet, but it's unlikely that much will change in core, hence why I'm PRing this now.

# How the hell does this work
### Segments
Fluid networks are modelled using 'segments', which are just collections of connected 'connection points'. For example, 3 machines might be connected together using pipes, and so all 3 of these machines are on the same segment. Not that pipes themselves are not implemented in this PR - connection points are abstract!

When a connection is made between two separate segments, they are merged together. When the connection is removed, they are split again. This is all handled by the `FluidManager` class, which also keeps track of all the points, segments, and tickers.

In the corresponding base PR (https://github.com/pylonmc/pylon-base/pull/50), I have added interactions that act as the corresponding physical connection point. This sounds complicated, but it's really not - see FluidConnectionInteraction in the base PR.

### Machines
Machines can implement the PylonFluidBlock interface to interface with the fluid system. It's important to note that this interface doesn't actually allow connecting pipes to the machine or anything - it just makes the machine accessible to the fluid ticker.

Machines can supply and request fluid. This is done by implementing the
`getSuppliedFluids(connectionPoint: String): Map<PylonFluid, Long>`
and
`getRequestedFluids(connectionPoint: String): Map<PylonFluid, Long>`
methods. When ticking a segment, the ticker will use this to determine what the supply and demand of fluid is, and route fluids accordingly. Segments can only route one fluid type per tick, and distribute fluid based on a round-robin system.

The ticker will then tell each machine how much fluid it has actually received or had removed using the
`addFluid(connectionPoint: String, fluid: PylonFluid, amount: Long)`
and
`removeFluid(connectionPoint: String, fluid: PylonFluid, amount: Long)`
methods. State (ie: how much fluid is contained, and of what type) must be kept track of by the implementer.

See the fluid tank in the corresponding base PR (https://github.com/pylonmc/pylon-base/pull/50) for a full example.

### PylonFluid
Fluid types are instances of PylonFluid. I added a fluids registry for this. Fluids can have 'tags' which is just some data (an arbitrary class) associated with a fluid, like temperature or whether the fluid is a metal. PylonFluid has methods for dealing with these tags.

### Flow rate & predicates
You might want to limit the flow rate of segments, or what fluids a segment allows through. You can do this by using FluidManager's `setFluidPerTick(Segment)` and `setFluidPredicate(segment)`. 

When two segments are merged, one of them is picked in an effectively random way, and the flow rate and predicate from that segment is used for the new segment. When a segment is split, the two new segments will inherit the flow rate and predicate from the old segment.

# Known issues / future work
I have taken some shortcuts in favour of getting this done faster because it is blocking smelting (and also there are other arguably higher priority projects ATM). There are several TODOs floating around where this is the case. Also:
- This system is likely to be slow to begin with. Optimisation will be needed once we have enough data, or a setup to test performance better
- The flow rate / predicate merging logic is a bit janky, but I'm not sure how to fix this


pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/config/PylonConfig.kt
```diff
@@ -7,13 +7,16 @@ object PylonConfig {
...
     @JvmStatic
-    val tickDelay: Int by config
+    val tickRate: Int by config
 
     @JvmStatic
     val allowedBlockErrors: Int by config
 
     @JvmStatic
-    val waliaInterval: Int by config
+    val waliaIntervalTicks: Int by config
```
Seggan, line 16: I just realized it's misspelled lol, it's supposed to be `waila`

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/datatypes/PylonFluidPersistentDataType.kt
```diff
@@ -0,0 +1,23 @@
+package io.github.pylonmc.pylon.core.datatypes
+
+import io.github.pylonmc.pylon.core.fluid.PylonFluid
+import io.github.pylonmc.pylon.core.registry.PylonRegistry
+import org.bukkit.persistence.PersistentDataAdapterContext
+import org.bukkit.persistence.PersistentDataType
+
+object PylonFluidPersistentDataType : PersistentDataType<String, PylonFluid> {
```
Seggan, line 8: Can be replaced with `PylonSerializers.KEYED`

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/fluid/FluidConnectionPoint.kt
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * A point is just a connection in a fluid network, like a machine's output or the end of a pipe.
+ *
+ * The FluidConnectionPoint class is created in memory for every loaded point. FluidConnectionPoints
+ * are persisted when unloaded, but do not store the segment UUID - this is decided at runtime.
+ *
+ * The flow rate of a segment is the lowest maxFlowRate of any point in the segment.
+ */
+class FluidConnectionPoint(
```
Seggan, line 16:
```suggestion
data class FluidConnectionPoint(
```
Nitpick

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/fluid/FluidManager.kt
```diff
@@ -0,0 +1,383 @@
...
+            tickers[point.segment]!!.cancel()
+        }
+    }
+
+    /**
+     * Call when creating a new connection point, or when one has been loaded
+     */
+    @JvmStatic
+    fun add(point: FluidConnectionPoint) {
+        check(!points.contains(point.id)) { "Duplicate connection point" }
```
Seggan, line 93:
```suggestion
        check(point.id !in points) { "Duplicate connection point" }
```
Kotlin convention

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/fluid/FluidManager.kt
```diff
@@ -0,0 +1,383 @@
...
+    /**
+     * Sets the flow rate per tick for a segment. The segment will not transfer more fluid than the
+     * flow rate per tick.
+     *
+     * Preserved across disconnects and connects (when connecting two points, one of the two points
+     * being connected is selected, and its segment's flow rate is copied to the new segment).
+     */
+    @JvmStatic
+    fun setFluidPerTick(segment: UUID, fluidPerTick: Long) {
+        check(segments.contains(segment)) { "Segment does not exist" }
```
Seggan, line 134:
```suggestion
        check(segment in segments) { "Segment does not exist" }
```
Kotlin convention

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/fluid/FluidManager.kt
```diff
@@ -0,0 +1,383 @@
...
+        segments[segment]!!.predicate = predicate
+    }
+
+    /**
+     * Connects two points - and all their connected points - into one segment
+     */
+    @JvmStatic
+    fun connect(point1: FluidConnectionPoint, point2: FluidConnectionPoint) {
+        check(segments.contains(point1.segment)) { "Attempt to connect a nonexistant segment" }
+        check(segments.contains(point2.segment)) { "Attempt to connect a nonexistant segment" }
```
Seggan, line 157: I'm tired of writing diffs lol, just replace all the `contains` with `in`/`!in`

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/fluid/FluidManager.kt
```diff
@@ -0,0 +1,383 @@
...
+        }
+    }
+
+    /**
+     * Recursively gets all the points connected to another point *that are loaded*
+     */
+    @JvmStatic
+    fun getAllConnected(point: FluidConnectionPoint): Set<FluidConnectionPoint> {
+        val visitedPoints: MutableSet<FluidConnectionPoint> = mutableSetOf()
+        val pointsToVisit: MutableSet<FluidConnectionPoint> = mutableSetOf(point)
```
Seggan, line 208: Why not list

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/fluid/FluidManager.kt
```diff
@@ -0,0 +1,383 @@
...
+        return suppliedFluids
+    }
+
+    @JvmStatic
+    fun getSuppliedFluids(segment: UUID): Map<PylonFluid, Set<FluidSupplier>> {
+        val suppliedFluids: MutableMap<PylonFluid, MutableSet<FluidSupplier>> = mutableMapOf()
+        for (point in getPoints(segment, FluidConnectionPoint.Type.OUTPUT)) {
+            for (supplier in getSuppliedFluids(point)) {
+                suppliedFluids.putIfAbsent(supplier.fluid, mutableSetOf())
+                suppliedFluids[supplier.fluid]!!.add(supplier)
```
Seggan, line 263:
```suggestion
                suppliedFluids.getOrPut(supplier.fluid, ::mutableSetOf).add(supplier)
```

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/fluid/FluidManager.kt
```diff
@@ -0,0 +1,383 @@
...
+        return requestedFluids
+    }
+
+    @JvmStatic
+    fun getRequestedFluids(segment: UUID): Map<PylonFluid, Set<FluidRequester>> {
+        val requestedFluids: MutableMap<PylonFluid, MutableSet<FluidRequester>> = mutableMapOf()
+        for (point in getPoints(segment, FluidConnectionPoint.Type.INPUT)) {
+            for (requester in getRequestedFluids(point)) {
+                requestedFluids.putIfAbsent(requester.fluid, mutableSetOf())
+                requestedFluids[requester.fluid]!!.add(requester)
```
Seggan, line 302:
```suggestion
                requestedFluids.getOrPut(requester.fluid, ::mutableSetOf).add(requester)
```

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/fluid/FluidManager.kt
```diff
@@ -0,0 +1,383 @@
...
+        val blockRequestedFluids: Map<PylonFluid, Long>
+        try {
+            if (!point.position.chunk.isLoaded) {
+                return setOf()
+            }
+            block = BlockStorage.getAs<PylonFluidBlock>(point.position) ?: return setOf()
+            blockRequestedFluids = block.getRequestedFluids(point.name)
+        } catch (t: Throwable) {
+            t.printStackTrace()
+            return setOf()
```
Seggan, line 283: Why silent fail?

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/fluid/PylonFluid.kt
```diff
@@ -0,0 +1,41 @@
...
+    private val tags: MutableList<PylonFluidTag>,
+) : Keyed {
+
+    constructor(key: NamespacedKey, material: Material, vararg tags: PylonFluidTag)
+            : this(key, material, mutableListOf(*tags))
+
+    override fun getKey(): NamespacedKey
+        = key
+
+    fun withTag(tag: PylonFluidTag): PylonFluid
```
Seggan, line 20: `withTag` makes me think it's returning a new fluid, just make `addTag` return `this`

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/fluid/PylonFluid.kt
```diff
@@ -0,0 +1,41 @@
...
+    fun addTag(tag: PylonFluidTag) {
+        check(!hasTag(tag.javaClass)) { "Fluid already has a tag of the same type" }
+        tags.add(tag)
+    }
+
+    fun hasTag(type: Class<out PylonFluidTag>): Boolean
+        = getTag(type) != null
+
+    fun <T: PylonFluidTag> getTag(type: Class<T>): T?
+        = type.cast(tags.firstOrNull { type.isInstance(it) })
```
Seggan, line 32: Add an inline reified version for us Kotlin people

pylon-core/src/main/resources/config.yml
```diff
@@ -1,11 +1,14 @@
 # The amount of Minecraft ticks between each Pylon tick
-tick-delay: 1
+tick-rate: 1
 
 # The number of errors a ticking block can throw before it is disabled
 allowed-block-errors: 5
 
 # The interval, in ticks, between WALIA checks
-walia-interval: 1
+walia-interval-ticks: 1
```
Seggan, line 8: Same here, `walia` is misspelled

Seggan (changes requested):
Solid

pylon-core/src/main/resources/config.yml
LordIdra: oh yeah lol I missed that completely

Seggan: no i think i did the classes correctly

LordIdra: I'm leaving this to another PR since this is Not My Problem:tm:

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/fluid/FluidManager.kt
```diff
@@ -0,0 +1,383 @@
...
+        return visitedPoints
+    }
+
+    @JvmStatic
+    fun getPoints(segment: UUID, type: FluidConnectionPoint.Type): List<FluidConnectionPoint>
+        = segments[segment]!!.points.filter { it.type == type }
+
+    data class FluidSupplier(val block: PylonFluidBlock, val name: String, val fluid: PylonFluid, val amount: Long)
+
+    data class FluidRequester(val block: PylonFluidBlock, val name: String, val fluid: PylonFluid, val amount: Long)
```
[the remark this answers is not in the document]
LordIdra: Yes, added a comment to explain

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/fluid/FluidManager.kt
```diff
@@ -0,0 +1,383 @@
...
+        val blockSuppliedFluids: Map<PylonFluid, Long>
+        try {
+            if (!point.position.chunk.isLoaded) {
+                return setOf()
+            }
+            block = BlockStorage.getAs<PylonFluidBlock>(point.position) ?: return setOf()
+            blockSuppliedFluids = block.getSuppliedFluids(point.name)
+        } catch (t: Throwable) {
+            t.printStackTrace()
+            return setOf()
```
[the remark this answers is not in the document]
LordIdra: Don't really want a whole segment to stop working if one machine on it is throwing errors. Just makes things a bit more robust, and an exception here might lead to inconsistent state.

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/fluid/FluidManager.kt
LordIdra: Same thing as above

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/datatypes/PylonSerializers.kt
```diff
@@ -81,6 +86,12 @@ object PylonSerializers {
     @JvmField
     val INVENTORY = InventoryPersistentDataType
 
+    @JvmField
+    val PYLON_FLUID = PylonFluidPersistentDataType
```
Seggan, line 90: Can just use `KEYED.keyedTypeFrom`

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/fluid/FluidManager.kt
```diff
@@ -0,0 +1,389 @@
...
+
+    /**
+     * Recursively gets all the points connected to another point *that are loaded*
+     */
+    @JvmStatic
+    fun getAllConnected(point: FluidConnectionPoint): Set<FluidConnectionPoint> {
+        val visitedPoints: MutableSet<FluidConnectionPoint> = mutableSetOf()
+        val pointsToVisit: MutableList<FluidConnectionPoint> = mutableListOf(point)
+        while (pointsToVisit.isNotEmpty()) {
+            // why does java/kotlin not have a Set.pop method???
```
Seggan, line 213: Unneeded comment now

Seggan (approved):

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/fluid/FluidManager.kt
```diff
@@ -0,0 +1,404 @@
...
+        }
+    }
+
+    private fun startTicker(segment: UUID) {
+        check(segment !in tickers) { "Ticker already active" }
+
+        val dispatcher = pluginInstance.minecraftDispatcher
+        tickers[segment] = pluginInstance.launch(dispatcher) {
+            while (true) {
+                delay(PylonConfig.fluidIntervalTicks.toLong())
```
Seggan, line 399:
```suggestion
                delay(PylonConfig.fluidIntervalTicks.ticks)
```
`delay` uses milliseconds, `.ticks` converts ticks to milliseconds

pylon-test/src/main/java/io/github/pylonmc/pylon/test/fluid/Fluids.java
```diff
@@ -0,0 +1,18 @@
...
+
+import io.github.pylonmc.pylon.core.fluid.PylonFluid;
+import io.github.pylonmc.pylon.test.PylonTest;
+import org.bukkit.Material;
+
+
+public class Fluids {
+
+    public static final PylonFluid WATER = new PylonFluid(PylonTest.key("water"),  Material.CYAN_CONCRETE);
+    public static final PylonFluid LAVA = new PylonFluid(PylonTest.key("lava"), Material.ORANGE_CONCRETE)
```
Seggan, line 11: Arguably these should be part of core

Seggan (changes requested):

pylon-core/src/main/resources/lang/pyloncore/en.yml
Seggan, line 1: Missing keys in enws

pylon-core/src/main/resources/lang/pyloncore/en.yml
```diff
@@ -39,4 +39,7 @@ quantity:
   seconds: "s"
   hearts: "❤"
   percent: "%"
-  research_points: "points"
\ No newline at end of file
+  research_points: "points"
+  temperature: "°C"
+  fluid: "mB"
+  fluid_per_second: "mB/s"
```
Seggan, line 45:
```suggestion
  research_points: " points"
  temperature: "°C"
  fluid: " mB"
  fluid_per_second: " mB/s"
```

Seggan (changes requested):

pylon-test/src/main/java/io/github/pylonmc/pylon/test/fluid/Fluids.java
LordIdra: I don't think so. They're 'content'. This would be the only 'content' in core.

Seggan:
Still missing enws

pylon-test/src/main/java/io/github/pylonmc/pylon/test/fluid/Fluids.java
Seggan: Phantom Block and the debug stairs are also content

LordIdra: Yea but they're inherent parts of core. Another implementation of base might want to not have water/lava. Not the same for phantom block / debug stairs. You could also argue in that case milk should be a liquid because it's part of minecraft itself so should be in core

Seggan: Fair

pylon-core/src/main/resources/lang/pyloncore/enws.yml
```diff
@@ -36,4 +36,7 @@ quantity:
   seconds: "s"
   hearts: "❤"
   percent: "%"
-  research_points: "points"
\ No newline at end of file
+  research_points: "points"
+  temperature: "°C"
+  fluid: "mB"
+  fluid_per_second: "mB/s"
```
Seggan, line 42:
```suggestion
  research_points: " points"
  temperature: "°C"
  fluid: " mB"
  fluid_per_second: " mB/s"
```

Seggan (approved):
One last bit

pylon-core/src/main/resources/lang/pyloncore/enws.yml
LordIdra: oops lol

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/fluid/FluidManager.kt
```diff
@@ -0,0 +1,405 @@
...
+            tickers[point.segment]!!.cancel()
+        }
+    }
+
+    /**
+     * Call when creating a new connection point, or when one has been loaded
+     */
+    @JvmStatic
+    fun add(point: FluidConnectionPoint) {
+        check(point.id !in points) { "Duplicate connection point" }
```
OhmV-IR, line 97: Most of these checks can become severe warns, don't need to throw imo since there's no return value so execution can reasonably continue

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/fluid/FluidManager.kt
```diff
@@ -0,0 +1,405 @@
...
+                }
+            }
+
+            // Only allow one type of fluid to be distributed per tick
+            break
+        }
+    }
+
+    private fun startTicker(segment: UUID) {
+        check(segment !in tickers) { "Ticker already active" }
```
OhmV-IR, line 395: something looks off with this logic... should be segment in tickers right? also this can probably just be an early return instead of a throw

OhmV-IR (approved):
LGTM, but my main complaint is that FluidManager throws exceptions out like candy on halloween

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/fluid/FluidManager.kt
LordIdra: I think this is fine, if we're trying to add connection points twice, it's very likely something is wrong with the underlying logic so I'd prefer a hard fail with a nice stacktrace here

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/fluid/FluidManager.kt
LordIdra: Nope - this check makes sure that a ticker hasn't *already* been started. I don't think it should be a early return - as with the other check you mentioned, if this is being called twice, something very very fishy is going on so I'd prefer a hard fail with a nice stacktrace

OhmV-IR (approved):
hardest rereview ever

PR #102, opened 2025-05-13
Pass events here too

Intybyte:
Event needs to be passed here too so addons/base can know what the cause of the explosion is when an entity explodes and breaks up blocks


pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/context/BlockBreakContext.kt
```diff
@@ -23,11 +24,11 @@ interface BlockBreakContext {
...
     // No event parameter since explosion can be either from block or entity
-    data object WasExploded : BlockBreakContext {
+    data class EntityExploded(val event: EntityExplodeEvent) : BlockBreakContext {
         override val normallyDrops = true
     }
 
-    data class Exploded(val event: BlockExplodeEvent) : BlockBreakContext {
+    data class BlockExploded(val event: BlockExplodeEvent) : BlockBreakContext {
         override val normallyDrops = false
     }
```
Seggan, line 33: There is a crucial difference between `Exploded` and `WasExploded`: the formaer means that the block itself is doing the explosion, while the latter means it was exploded either by another block or entity. So this code doesn't work

Seggan (approved):

LordIdra (approved):

PR #103, opened 2025-05-13
Fix block placing inside entity

Intybyte:
you can place blocks inside players, we don't want that


Intybyte:
probably we need to check for entities in general in said blocks location too... will test later

Seggan (approved):

Intybyte:
Tested

OhmV-IR (changes requested):
ig test went fine, just one code cleanup comment. I want test for pylonEntities in though before merge

Seggan:
This will be replaced by a rework of the block placing system, see https://discord.com/channels/1329177304857055273/1329177972631928915/1371989721047236658

Issue #104, opened 2025-05-13
breaking pylon blocks does not take item durability


glomdom:
when breaking a pylon block in survival, the used tool does not take durability

PR #105, opened 2025-05-13
damage tools when breaking pylon blocks (closes: #104)

glomdom:
closes: #104

see issue for details; this fixes it

might not be an optimal solution

### DIFF
--- a/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/BlockListener.kt
+++ b/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/BlockListener.kt
@@ -26,7 +26,7 @@ import org.bukkit.event.inventory.FurnaceExtractEvent
 import org.bukkit.event.player.PlayerTakeLecternBookEvent
 import org.bukkit.event.player.PlayerInteractEvent
 import org.bukkit.inventory.meta.Damageable
-import java.util.Random
+import kotlin.random.Random
 
 /**
  * This listener listens for various events that would indicate a Pylon block either
@@ -60,7 +60,7 @@ internal object BlockListener : Listener {
 
             val unbreakingLevel = tool.getEnchantmentLevel(Enchantment.UNBREAKING)
             val shouldDamage = if (unbreakingLevel > 0) {
-                Random().nextInt(unbreakingLevel + 1) == 0
+                Random.Default.nextInt(unbreakingLevel + 1) == 0
             } else true
 
             if (shouldDamage) {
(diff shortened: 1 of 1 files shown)


pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/BlockListener.kt
```diff
@@ -47,6 +51,27 @@ internal object BlockListener : Listener {
...
+
+            if (shouldDamage) {
+                meta.damage += 1
+                if (meta.damage >= tool.type.maxDurability) {
+                    player.inventory.setItemInMainHand(null)
+                    player.playSound(player.location, Sound.ENTITY_ITEM_BREAK, 1.0f, 1.0f)
+                } else {
+                    tool.itemMeta = meta
+                }
+            }
```
Seggan, line 74: Instead of all this you can just do `tool.damage(1, player)`. It does all the calculations too

Seggan (changes requested):

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/BlockListener.kt
glomdom: resolved in b59701302ab270f9a1ba620e1939491956381f7d

Seggan (approved):

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/BlockListener.kt
```diff
@@ -47,6 +48,11 @@ internal object BlockListener : Listener {
         if (BlockStorage.isPylonBlock(event.block)) {
             BlockStorage.breakBlock(event.block, BlockBreakContext.PlayerBreak(event))
             event.isDropItems = false
+
+            val player = event.player
+            val tool = player.inventory.itemInMainHand
+
+            tool.damage(1, player)
```
OhmV-IR, line 55: you need to be calling this event here: https://jd.papermc.io/paper/1.21.5/org/bukkit/event/player/PlayerItemDamageEvent.html, note that "damage" in the constructor means its new hp

OhmV-IR (changes requested):
Works fine in-game, but is missing an event call

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/BlockListener.kt
Seggan: Hmm, I thought damage would auto call the event?

Seggan:
This is related to breaking, not placing blocks though?

LordIdra:
Oh, sorry, got this mixed up with the other PR

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/BlockListener.kt
Seggan: Looked through NMS code, can confirm it calls the event

OhmV-IR (approved):
LGTM CW&IG

PR #126, opened 2025-05-30
Add language file warnings

LordIdra:
Adds some warnings for translations
- Checks that the name and lore of a PylonItem is a translation key
- Checks that name and lore keys exist in all language files for each item
- Checks that blocks have a WAILA entry (I had to very slightly change WailaConfig for this)


pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/PylonBlock.kt
```diff
@@ -67,11 +68,26 @@ open class PylonBlock protected constructor(val block: Block) {
...
+        fun register(schema: PylonBlockSchema) {
+            val translator = AddonTranslator.translators[schema.addon]
+            check(translator != null) {
+                "Addon does not have a translator; did you forget to call registerWithPylon()?"
+            }
+
+            for (locale in schema.addon.languages) {
+                val wailaKey = WailaConfig.getWailaTranslationKey(schema.key)
+                if (!translator.translationKeyExists(wailaKey, locale)) {
+                    PylonCore.logger.warning("Block ${schema.key} is missing a WAILA translation key (${locale.displayName} | $wailaKey")
```
Seggan, line 80: What if I want a custom translation key (this actually happens in smeltery)

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/item/PylonItem.kt
```diff
@@ -34,15 +38,48 @@ open class PylonItem(val stack: ItemStack) : Keyed {
...
+            var isNameAndLoreValid = true
+            if (name != ItemStackBuilder.nameKey(schema.key)) {
+                PylonCore.logger.warning("Item ${schema.key}'s name is not a translation key; check your item uses ItemStackBuilder.pylonItem(...)")
+                isNameAndLoreValid = false
+            }
+
+            if (lore == null || lore != ItemStackBuilder.loreKey(schema.key)) {
+                PylonCore.logger.warning("Item ${schema.key}'s lore is not a translation key; check your item uses ItemStackBuilder.pylonItem(...)")
+                isNameAndLoreValid = false
+            }
```
Seggan, line 60: Why are you extracting the text just do `is TranslatableComponent`

Seggan (changes requested):

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/PylonBlock.kt
LordIdra: Oh that complicates things. What did you need it for in smeltery, and could it be emulated by just using a single placeholder as the waila text (like `smeltery: %text%`)? It would be very nice to have error checking here and we can't really do that with custom keys

LordIdra: ( @Seggan not sure if you missed this )

OhmV-IR (changes requested):
Did not pass IG vibe test, it didn't warn for missing waila keys, tested with latest base master as of writing. Also I don't think checking that every itemstack's name is a translation key is necessary because it throws earlier if the stack isnt a pylonitem which happens if its generated from ItemStackBuilder.of or smth like that. 
See error:
image

LordIdra:
> Did not pass IG vibe test, it didn't warn for missing waila keys, tested with latest base master as of writing. Also I don't think checking that every itemstack's name is a translation key is necessary because it throws earlier if the stack isnt a pylonitem which happens if its generated from ItemStackBuilder.of or smth like that. See error: image

The WAILA stuff was removed, see above convo with seggan. I just didn't update PR description. 

Regarding checking that it's a translation key - it will throw anyway if ItemStackBuilder.of or something like that is used. The idea is to make sure people are using ItemStackBuilder.pylonItem and then providing translation keys in the lang file rather than hardcoding in strings

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/PylonBlock.kt
Seggan: It works too, but I would still prefer a method to disable such warnings for specific items/blocks because of, as I've mentioned often, generated items.

LordIdra: oh, ffs yeah I completely forgot about that

PR #135, opened 2025-06-08
New arrow events

OhmV-IR:
Adds onArrowHit and onArrowDamage, helps with items wanting to make more custom effects, eg the ice arrow


pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/entity/EntityListener.kt
```diff
@@ -33,4 +38,24 @@ object EntityListener : Listener {
             event.pylonEntity.onDeath(event)
         }
     }
+
+    @EventHandler
+    fun handle(event: ProjectileHitEvent) {
+        if(event.entity is AbstractArrow){
```
Seggan, line 44: why is it abstract arrow

OhmV-IR: there's a lot of arrow classes, eg SpectralTippedArrow, Arrow, TippedArrow .etc AbstractArrow is the top-level iface that provides the getItemStack() method that we're using here

OhmV-IR:
base is meant to be master, not merging this yet. I'll probably have to add more things to it for the add-weapons stuff.

Seggan:
Without the base not being master, it shows all iface changes too. WHen you merge the other pr the base will auto change

OhmV-IR:
k then didn't know that

LordIdra (changes requested):
Needs formatting, then happy

Seggan (approved):

LordIdra (approved):

PR #138, opened 2025-06-13
Jump and Sneak events include inside player's location

ihateblueb:
Fixes problem with elevator pr where the elevator doesn't work as a bottom slab.


LordIdra (approved):
I assume you tested

ihateblueb:
Yup, always

Seggan (approved):
ngl I wouldn't have thought to use the elvis operator like that, great job :+1:

PR #139, opened 2025-06-14
add PylonBlock waila fallback to name

ihateblueb:
Reduces duplicate translation values where names were the same as waila


pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/PylonBlock.kt
```diff
@@ -32,7 +31,11 @@ open class PylonBlock protected constructor(val block: Block) {
     @JvmSynthetic
     internal var errorBlock: BlockDisplay? = null
 
-    open val name: Component = Component.translatable("pylon.${schema.key.namespace}.item.${schema.key.key}.waila")
+    open val name: Component = Component.translatable("pylon.${schema.key.namespace}.item.${schema.key.key}.waila").let {
+        if (it.children().isEmpty())
```
Seggan, line 35: Im confused on how this code works. How is it checking that the waila key does not exist? The translation key will always have 0 children

Seggan (approved):

LordIdra (approved):

PR #158, opened 2025-07-01
Hydraulic machines fixes

LordIdra:
Just a few small fixes I made while working on hydraulic machines
- fixed pylon guide search feature breaking if any item/fluid does not have a valid name
- fix not being able to sneak to place block on a PylonGuiBlock (I am not sure if my changes will break something - @Seggan if you could have a look that would be good)
- renamed Component to MultiblockComponent
- added getItem method to config section which will get a pylon item or a material


Seggan:
Also it seems to have broken opening GUIs which would be normally a different GUI. ie smeltery is opening blast furnace GUI, not smeltery GUI

PR #160, opened 2025-07-03
Restore faith in CI

LordIdra:
- Fix unregister not being called on main thread in a test
- Deletes the fluid loop test. I investigated this in-game myself and discovered that loop behaviour is non-deterministic by nature of the implementation. Taking fluids from machines is done on a first-come-first-serve basis - so if your producer is producing enough to saturate the loop, it may be that one of the blocks in the loop just takes all its fluid from the producer and none from the second block. Yeah it's unintuitive but probably not worth the time and performance hit to fix

closes https://github.com/pylonmc/pylon-core/issues/159


Seggan (approved):

OhmV-IR (approved):
LGTM

PR #164, opened 2025-07-07
Fix vanilla translations

### LINKED ISSUE #152: Fix vanilla item names not being wrapped properly
`ItemStack#effectiveName()` returns a translatable component for vanilla items, and since those are translated clientside, we cannot do serverside translation for that. The solution would be to download the client translations ([which I've done before](https://github.com/Seggan/SpigotWorldGenVisualizer/blob/97f25184994fc24f51929f28f486f062fba06863/app/src/main/kotlin/io.github.seggan.swgv/App.kt#L54)) and manually hook those into the global translator so we can translate minecraft stuff serverside too.

Seggan:
Resolves #152

### DIFF
--- a/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/i18n/MinecraftTranslator.kt
+++ b/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/i18n/MinecraftTranslator.kt
@@ -13,6 +13,11 @@ import java.text.MessageFormat
 import java.util.Locale
 import kotlin.io.path.*
 
+/**
+ * Since Pylon has automatic line wrapping for items, it needs to know the length of the translated strings.
+ * However, this is not possible with Minecraft translation keys since that's done client-side.
+ * This class downloads the Minecraft language files and uses them to translate some keys server-side.
+ */
 internal object MinecraftTranslator : Translator {
 
     private const val LAUNCHER_META_URL = "https://launchermeta.mojang.com/mc/game/version_manifest_v2.json"
--- a/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/i18n/MinecraftTranslator.kt
+++ b/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/i18n/MinecraftTranslator.kt
@@ -1,0 +1,97 @@
+package io.github.pylonmc.pylon.core.i18n
+
+import com.google.gson.JsonObject
+import com.google.gson.JsonParser
+import io.github.pylonmc.pylon.core.PylonCore
+import io.github.pylonmc.pylon.core.util.pylonKey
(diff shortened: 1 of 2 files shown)


pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/i18n/MinecraftTranslator.kt
```diff
@@ -0,0 +1,97 @@
...
+internal object MinecraftTranslator : Translator {
+
+    private const val LAUNCHER_META_URL = "https://launchermeta.mojang.com/mc/game/version_manifest_v2.json"
+    private val LANG_REGEX = """^minecraft/lang/(.+)\.json$""".toRegex()
+
+    private val cacheDir = Path("cache/pylon/mclang")
+    private val translations = mutableMapOf<Locale, Map<String, MessageFormat>>()
+
+    init {
+        PylonCore.logger.info("Loading vanilla translations, this may take a while...")
```
LordIdra, line 25: Tiny thing but maybe 'but only needs to be done on the first run of Pylon'

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/i18n/MinecraftTranslator.kt
```diff
@@ -0,0 +1,97 @@
...
+import java.net.URI
+import java.nio.file.FileSystems
+import java.text.MessageFormat
+import java.util.Locale
+import kotlin.io.path.*
+
+internal object MinecraftTranslator : Translator {
+
+    private const val LAUNCHER_META_URL = "https://launchermeta.mojang.com/mc/game/version_manifest_v2.json"
+    private val LANG_REGEX = """^minecraft/lang/(.+)\.json$""".toRegex()
```
LordIdra, line 19: What the fuck

LordIdra (commented):
nice work, but think you couldadd a quick explanation at the top of MinecraftTranslator? it's very unclear to me what the file does at first glance. otherwise happy

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/i18n/MinecraftTranslator.kt
Seggan: ? it's a regex for matching and extracting the lang files form the assets index

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/i18n/MinecraftTranslator.kt
LordIdra: Oh my bad

LordIdra (approved):

PR #165, opened 2025-07-07
Fix WAILA not appearing sometimes

Seggan:


LordIdra:
Ohhh wait I didn't realise commit messages counted, that's cool

PR #172, opened 2025-07-13
add meters per second UnitFormat

OhmV-IR:


LordIdra (changes requested):
Think we could do blocks per second? Since we already have a blocks unit but not metres unit, and makes more sense in context of the game

LordIdra (approved):
I think b/s is fine, it'll be obvious from context

PR #184, opened 2025-07-22
Replace createEntities with addEntity

LordIdra:
Removes `createEntities` and replaces it with `addEntity` which can be called in the constructor.

This is nice because is removes some boilerplate, helps make sure anything related to creating the block is done in the place constructor, and allows entities to be accessed from the place constructor.


OhmV-IR (approved):
LGTM

PR #188, opened 2025-07-24
Replace createEntities with addEntity

LordIdra:
Removes `createEntities` and replaces it with `addEntity` which can be called in the constructor.

This is nice because is removes some boilerplate, helps make sure anything related to creating the block is done in the place constructor, and allows entities to be accessed from the place constructor.

See https://github.com/pylonmc/pylon-base/pull/104

### DIFF
--- a/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/base/PylonEntityHolderBlock.kt
+++ b/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/base/PylonEntityHolderBlock.kt
@@ -1,11 +1,9 @@
 package io.github.pylonmc.pylon.core.block.base
 
-import io.github.pylonmc.pylon.core.block.context.BlockCreateContext
 import io.github.pylonmc.pylon.core.datatypes.PylonSerializers
 import io.github.pylonmc.pylon.core.entity.EntityStorage
 import io.github.pylonmc.pylon.core.entity.PylonEntity
 import io.github.pylonmc.pylon.core.event.PylonBlockDeserializeEvent
-import io.github.pylonmc.pylon.core.event.PylonBlockPlaceEvent
 import io.github.pylonmc.pylon.core.event.PylonBlockSerializeEvent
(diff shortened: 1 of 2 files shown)


OhmV-IR (approved):
LGTM

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/base/PylonEntityHolderBlock.kt
```diff
@@ -25,11 +22,16 @@ import java.util.UUID
...
 interface PylonEntityHolderBlock : PylonBreakHandler {
 
-    fun createEntities(context: BlockCreateContext): Map<String, PylonEntity<*>> = mutableMapOf()
-
     @get:ApiStatus.NonExtendable
     val heldEntities: MutableMap<String, UUID>
-        get() = holders[this] ?: error("You cannot access held entities before the block is placed")
+        get() = holders.getOrPut(this) { mutableMapOf() }
+
+    fun addEntity(name: String, entity: PylonEntity<*>) {
```
Seggan, line 29: add nonextendable on this

Seggan (approved):

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/base/PylonEntityHolderBlock.kt
LordIdra: What's the reason for that?

LordIdra: Well maybe you're adding aircraft and want to add them all as passengers to a pig. Base already does some 'weird' stuff with extending interface methods like this which I wasn't expecting (see fluid filter). I don't think it's a good idea to mark them as nonextendable because technically they _can_ be extended just find

LordIdra: Yeah ohm beat me to it pretty much

Issue #142, opened 2025-06-19
Make multiblock ghosts visually distinct from normal blocks


LordIdra:
this is a problem for multiblocks like the grindstone. making them glow is probably the best way to do this.

PR #189, opened 2025-07-24
Multiblock improvements

LordIdra:
- Fix some issues with PylonEntityHolderBlock where the regular get methods would throw if the entity didn't exist
- Adds a 'facing' option to PylonSimpleMultiblock. If set, this sets the direction we expect the multiblock to be built in. This will be used both for checking the multiblock is formed, and for spawning ghost blocks
- Adds a glowing effect to multiblock ghosts - red if the block is incorrect, green if the block is correct.
- Abstracts out a few small methods
- Fixes an earlier mistake in the new fluid manager logic

closes https://github.com/pylonmc/pylon-core/issues/142


pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/base/PylonFluidBufferBlock.kt
```diff
@@ -80,8 +81,8 @@ interface PylonFluidBufferBlock : PylonFluidBlock {
      * than it technically has capacity for.
      */
     fun setFluidCapacity(fluid: PylonFluid, capacity: Double) {
-        check(capacity > 0)
-        fluidData(fluid).capacity = capacity
+        check(capacity > -1.0e6)
```
Seggan, line 84: This number seems to be used a lot, should probably make it a constant somewhere

Seggan (approved):

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/base/PylonFluidBufferBlock.kt
LordIdra: https://github.com/pylonmc/pylon-core/issues/202

PR #196, opened 2025-07-27
Replace get methods with setters and persistent data for block tickers

LordIdra:
- Replace isAsync and getCustomTickRate with setAsync and setTickRate
- Fix SimpleMultiblock storing data under fluid_tank_data key
- Standardise ticking settings to tick-interval (we were using several different names for this previously)

### DIFF
--- a/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/base/PylonTickingBlock.kt
+++ b/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/base/PylonTickingBlock.kt
@@ -30,6 +30,12 @@ interface PylonTickingBlock {
         tickingData.tickInterval = tickInterval
     }
 
+    /**
+     * WARNING: Settings a block to tick asynchronously could have unintended consequences.
+     *
+     * Only set this option if you understand what 'asynchronous' means, and note that you
+     * cannot interact with the world asynchronously.
+     */
     fun setAsync(isAsync: Boolean) {
         tickingData.isAsync = isAsync
     }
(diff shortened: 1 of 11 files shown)


pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/base/PylonTickingBlock.kt
```diff
@@ -1,11 +1,76 @@
...
+        get() = tickingData.isAsync
 
-    fun getCustomTickRate(globalTickRate: Int): Int = globalTickRate
+    fun setTickInterval(tickInterval: Int) {
+        tickingData.tickInterval = tickInterval
+    }
+
+    fun setAsync(isAsync: Boolean) {
+        tickingData.isAsync = isAsync
+    }
```
OhmV-IR, line 35: maybe add a warning doc too about how making something that should be synchronous async will cause all sorts of funny things to happen that you will *love*. I would consider making private too, but I think that is a step too far in the "policing addons" direction

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/datatypes/TickingBlockPersistentDataType.kt
OhmV-IR, line 1: Can we get a new PDC test for this class?

OhmV-IR (changes requested):

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/datatypes/TickingBlockPersistentDataType.kt
LordIdra: I think it's fine to leave it, haven't done this for any other recent pdts, the main reason we had so many in the past was to test with PylonPDC, but it's not really necessary to test them now because they're so simple imo

Seggan:
Why?

LordIdra:
> Why?

consistency (the other interfaces work like this now), marginal debugging advantage (you can see ticking rate and is async in pdc data), removes boilerplate (all the function code to return true) - it's a fairly small thing, just thought I may as well

PR #199, opened 2025-08-01
Allow silent block placement/breakment

Seggan:
Needed for the primitive smelting solution I came up with


OhmV-IR (approved):
It's okay, but I think this is going to add a lot of confusion and the block type assumption after block break / create is pretty common imo so I'd like to wait for @LordIdra 's opinion before approving

PR #200, opened 2025-08-05
Random stuff (again)

### LINKED ISSUE #150: Add DisplayRecipeType
Add DisplayRecipeType used for blocks like water pump, drainers, etc, which you can only add recipes to inside the class (no public addRecipe/removeRecipe/getRecipe methods). Then add display recipes to the relevant items in base

Seggan:
I promise this won't be another mega-pr

- Added silent block placement/breakment
- Added virtual recipe (closes #150)
- Added a util for formatting time
- Added an unclickable inventory for recipe display purposes
- Added options in `VanillaMultiblockComponent` for multiple types

### DIFF
--- a/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/PhantomBlock.kt
+++ b/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/PhantomBlock.kt
@@ -36,21 +36,22 @@ class PhantomBlock(
 ) : PylonBlock(block) {
 
     // Hacky placeholder
+    @Suppress("unused")
     internal constructor(block: Block, context: BlockCreateContext)
             : this(block.chunk.persistentDataContainer.adapterContext.newPersistentDataContainer(), pylonKey("bruh"), block) {
         throw UnsupportedOperationException("Phantom block cannot be placed")
     }
 
     // Hacky placeholder
+    @Suppress("unused")
     internal constructor(block: Block, pdc: PersistentDataContainer)
             : this(block.chunk.persistentDataContainer.adapterContext.newPersistentDataContainer(), pylonKey("bruh"), block) {
         throw UnsupportedOperationException("Phantom block cannot be loaded")
     }
 
-    override fun getWaila(player: Player): WailaConfig {
+    override fun getWaila(player: Player): WailaConfig? {
         return WailaConfig(
-            text = name,
-            placeholders = listOf(PylonArgument.of("block", erroredBlockKey.toString())),
+            text = defaultTranslationKey.arguments(PylonArgument.of("block", erroredBlockKey.toString())),
             color = BossBar.Color.RED
         )
     }
--- a/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/PylonBlock.kt
+++ b/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/PylonBlock.kt
@@ -49,10 +49,10 @@ open class PylonBlock protected constructor(val block: Block) {
     @JvmSynthetic
     internal var errorBlock: BlockDisplay? = null
 
-    /**
-     * The name is used for WAILA, and as the title of the block's GUI, should it have one
-     */
-    open val name: Component = Component.translatable("pylon.${schema.key.namespace}.item.${schema.key.key}.waila", "pylon.${schema.key.namespace}.item.${schema.key.key}.name")
+    val defaultTranslationKey = Component.translatable(
+        "pylon.${schema.key.namespace}.item.${schema.key.key}.waila",
+        "pylon.${schema.key.namespace}.item.${schema.key.key}.name"
+    )
 
     constructor(block: Block, context: BlockCreateContext) : this(block)
     constructor(block: Block, pdc: PersistentDataContainer) : this(block)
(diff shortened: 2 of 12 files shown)


OhmV-IR (changes requested):
one structural comment, didn't IG test yet

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/util/gui/unit/UnitFormat.kt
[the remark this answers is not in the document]
OhmV-IR: alright if it's too much of a pain, then it's not worth it

OhmV-IR (approved):
LGTM

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/context/BlockBreakContext.kt
```diff
@@ -21,12 +21,21 @@ interface BlockBreakContext {
     @get:JvmName("normallyDrops")
     val normallyDrops: Boolean
 
+    /**
+     * If true, the block will be set to air after the Pylon data is removed
+     */
+    @Suppress("INAPPLICABLE_JVM_NAME")
```
LordIdra, line 27: Uhhh what's this about? That seems important

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/recipe/PylonVirtualRecipe.kt
```diff
@@ -0,0 +1,23 @@
...
+import xyz.xenondevs.invui.gui.Gui
+
+/**
+ * A sort of "pseudo recipe" that does not correspond to any real recipe, but is used for displaying
+ * stuff in the guide. For example, the water pump uses a `PylonVirtualRecipe` to display the
+ * output of the pump in the guide, even though it's not a "recipe" per se.
+ *
+ * @see [io.github.pylonmc.pylon.core.content.guide.PylonGuide.addVirtualRecipe]
+ */
+interface PylonVirtualRecipe : Keyed {
```
LordIdra, line 15: What's the reason to separate out virtual recipes? It seems to me that there's no actual consequence of just using normal recipes instead of the new virtual recipes. Like for the water pump example, couldn't you just add a normal recipe? I don't see any functional difference between that and this

LordIdra (changes requested):

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/context/BlockBreakContext.kt
Seggan: `@JvmName` is not *fully* implemented for interface members, but only for pre-Java 8 targets (something something default interface members something something), but it's fine to just suppress the error if you're targeting Java 8 and later

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/recipe/PylonVirtualRecipe.kt
Seggan: Because you can add stuff to a normal recipe. Not to a virtual recipe (ex what would adding a recipe to the water pump even mean?). It's kinda weird to have a recipe type you can't add to, so better to just have a whole separate thing to make it clear in the code.

LordIdra: wouldn't you need to add recipes anyway..? Like you'd need a default 'produce 1000 water' recipe for the fluid pump

Seggan: No? Cause you just hardcode that. Doing it within the recipe system would be a lot harder. The problem here is *external* manipulation of recipes, not internal. There's also the possibility that recipes are inherited from an external source, such as the fluid strainer (back when it could only strain slurries), or the pit kiln (originally just followed the smeltery recipes as long as it was <= 1100C).

https://discord.com/channels/1329177304857055273/1329177972631928915/1385728490355298436

LordIdra: Oh I see what you mean with regards to default recipe. This is potentially going to create large problems down the line though especially with guide indexing - because we'll end up with a ton of duplicate logic everywhere. Would it perhaps be better here to do similar to what I suggested in the message you linked and have a SingletonRecipeType, which has one default recipe, that calls methods on the singleton recipe? So you'd actually have the SingletonRecipeType with all the abstract Recipe methods, and then the default recipe would just call those methods?

Seggan: I'm confused

LordIdra: pseudocode
```
class SingletonRecipeType : RecipeType<SingletonRecipe> {
   class SingletonRecipe : Recipe {
     override isInput() { return isInput() }
     ...
   }
   
   recipes = List.of(SingletonRecipe())
   
   isInput() { return ... }
}
```
So basically make use of non-static classes so implementors can treat the recipe type as a recipe. (I _think_ this should work?)

Seggan: There's still the problem of a spurious `addRecipe`

LordIdra: Or maybe not with the system I've outlined above. Maybe just throw IllegalArgumentException - a bit annoying but unlikely to ever be an issue, and less annoying than writing all the duplicate logic

LordIdra: Why not? It seems perfectly natural to me. I don't think a global recipe type would work because of the generic argument. I mean maybe. If you can make it work I'd be fine with that

PR #201, opened 2025-08-06
More WAILA stuff

Seggan:
Some WAILA stuff that I both needed for pit kiln and some old code cleanup after the translation cleanup.

### DIFF
--- a/pylon-core/build.gradle.kts
+++ b/pylon-core/build.gradle.kts
@@ -18,9 +18,6 @@ repositories {
     maven("https://repo.xenondevs.xyz/releases") {
         name = "InvUI"
     }
-    maven("https://repo.aikar.co/content/groups/aikar/") {
-        name = "Aikar"
-    }
 }
 
 dependencies {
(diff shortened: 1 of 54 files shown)


LordIdra (commented):
Do you think you could give more of a description of what the changes here do and why you made them? I'm looking through the code and just a bit confused on what's actually being changed here / why. (Especially changing WailaConfig so you have to pass arguments in getName - that seems like a step backwards to me)

Seggan:
The changing of `WailaConfig` is for consistency with literally every other place where we pass arguments into the translation component itself instead of some separate arguments field, back when `PylonArgument` wasn't actually a `TranslationArgument[Like]`

LordIdra:
Ah makes sense. Could we change getName to just return the name then, and have the convention be to use `Component.translateable(getName(), PylonArgument.of(...))` for parity with placeholders then?

Seggan:
Wow, I thought you'd be a lot harder to convince than that :sweat_smile:

OhmV-IR (changes requested):
re: many base bugs

PR #209, opened 2025-08-10
Load external langs

### LINKED ISSUE #208: Add 'language not supported' text in place of the translation key and stop spamming the console if language file not found
See https://github.com/pylonmc/pylon-core/issues/204

Seggan:
Makes the behavior as originally intended and allows loading language files put in `/lang/<addon>`. Also fixes #208.

### DIFF
--- a/nms/src/main/kotlin/io/github/pylonmc/pylon/core/i18n/packet/PlayerPacketHandler.kt
+++ b/nms/src/main/kotlin/io/github/pylonmc/pylon/core/i18n/packet/PlayerPacketHandler.kt
@@ -47,6 +47,7 @@ class PlayerPacketHandler(private val player: ServerPlayer, private val handler:
 
     private inner class PacketHandler : ChannelDuplexHandler() {
         override fun write(ctx: ChannelHandlerContext, packet: Any, promise: ChannelPromise) {
+            var packet = packet
             when (packet) {
                 is ClientboundContainerSetContentPacket -> {
                     packet.items.forEach(::translateItem)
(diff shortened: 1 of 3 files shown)


Seggan:
Ah I misread the issue

nms/src/main/kotlin/io/github/pylonmc/pylon/core/i18n/packet/PlayerPacketHandler.kt
```diff
@@ -47,15 +47,39 @@ class PlayerPacketHandler(private val player: ServerPlayer, private val handler:
 
     private inner class PacketHandler : ChannelDuplexHandler() {
         override fun write(ctx: ChannelHandlerContext, packet: Any, promise: ChannelPromise) {
+            var packet = packet
```
LordIdra, line 50: lol

LordIdra (approved):

nms/src/main/kotlin/io/github/pylonmc/pylon/core/i18n/packet/PlayerPacketHandler.kt
Seggan: In NMS, no crimes are crimes :P

OhmV-IR (changes requested):
oop actually can't get the server to use my en_CA.yml file after setting my locale in mc to English (Canada), and restarting server a few times, double checked file was in the same spot as the en.yml file too

Seggan:
Fixed

PR #226, opened 2025-08-14
Update 1.21.8

Seggan:
New features:
- Happy Ghast
- LA LA LA LAVAAAAAAA
- New item components
- ***DIALOGS***

### DIFF
--- a/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/util/gui/ProgressItem.kt
+++ b/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/util/gui/ProgressItem.kt
@@ -4,6 +4,7 @@ import io.github.pylonmc.pylon.core.i18n.PylonArgument
 import io.github.pylonmc.pylon.core.item.builder.ItemStackBuilder
 import io.github.pylonmc.pylon.core.util.gui.unit.UnitFormat
 import io.papermc.paper.datacomponent.DataComponentTypes
+import io.papermc.paper.datacomponent.item.TooltipDisplay
 import net.kyori.adventure.text.Component
 import org.bukkit.Material
 import org.bukkit.entity.Player
@@ -39,6 +40,10 @@ abstract class ProgressItem @JvmOverloads constructor(
             .set(DataComponentTypes.MAX_STACK_SIZE, 1)
             .set(DataComponentTypes.MAX_DAMAGE, MAX_DURABILITY)
             .set(DataComponentTypes.DAMAGE, (progressValue * MAX_DURABILITY).toInt())
+            .set(
+                DataComponentTypes.TOOLTIP_DISPLAY, TooltipDisplay.tooltipDisplay()
+                    .addHiddenComponents(DataComponentTypes.DAMAGE, DataComponentTypes.MAX_DAMAGE)
+            )
         totalTime?.let {
             val remaining = it - it * progress
             builder.lore(


nms/src/main/kotlin/io/github/pylonmc/pylon/core/i18n/packet/PlayerPacketHandler.kt
```diff
@@ -85,8 +86,22 @@ class PlayerPacketHandler(private val player: ServerPlayer, private val handler:
...
+                    val isEmpty = packet.carriedItem == HashedStack.EMPTY
+                    // force server to resend the item
+                    packet = ServerboundContainerClickPacket(
+                        packet.containerId,
+                        if (isEmpty) -1 else packet.stateId,
+                        packet.slotNum,
+                        packet.buttonNum,
+                        packet.clickType,
+                        packet.changedSlots,
+                        if (isEmpty) packet.carriedItem else HashedStack.EMPTY,
```
OhmV-IR, line 101: if it's empty can you not just leave the packet as is? its nms though so I won't question it too much I guess?

settings.gradle.kts
```diff
@@ -12,4 +12,3 @@ plugins {
 include("pylon-core")
 include("pylon-test")
 include("nms")
-include("dokka-plugin")
```
OhmV-IR, line 15: ? pretty sure we're still using this no?

OhmV-IR (changes requested):
image
I think the nms is borked. And yes, I did make sure I was using 1.21.8
With a fresh clone of the master project and core/base too.

Seggan:
> image
> I think the nms is borked. And yes, I did make sure I was using 1.21.8
> With a fresh clone of the master project and core/base too.

Are you using a 1.21.8 *client*?

nms/src/main/kotlin/io/github/pylonmc/pylon/core/i18n/packet/PlayerPacketHandler.kt
Seggan: It's complicated, basically when it's empty is when you place an item into an empty slot, but since it's empty the server just oks it and doesn't send the set slot packet, so we have to resend the whole inv

Lagcatalyst:
> * _**DIALOGS**_
I probably shouldn't be here, buttttttttttttt
people have been abusing these to make ui, all I'm saying

settings.gradle.kts
Seggan: This is a thing from way back when when I was trying out a custom dokka plugin, I just forgot to remove this piece of config

Seggan:
I am unable to reproduce

OhmV-IR:
Alright here's exactly what I did, line by line in cmd:
1. rmdir /s pylon
2. Clone master proj
3. Clone base and core, rename appropriately
4. Gradlew runSnapshotServer
5. Launch mc 1.21.8
6. Try joining 127.0.0.1
I no longer have access to pc though so idra will have to test & review this one

Seggan:
1. You don't seem to be switching to the appropriate branches
2. You don't have to clone base and core at all? Just run `./gradlew` and it will autoclone

LordIdra (changes requested):
image
WAILA borked as mentioned on Discord

OhmV-IR (commented):
Looks fine cw then

PR #228, opened 2025-08-15
Fix 222 with enhancements

### LINKED ISSUE #222: BlockListener cancelled specify BlockPlaceEvent accidentally
When I holds a Building Wand (my custom PylonItem) to trigger a FakeBlockPlaceEvent to check permission, the event was cancelled by BlockListener accidentally.

Image

Source Code:

Image

In this code, the Building Wand is never a placeable PylonItem. `pylonBlock` in this case is always null, the event always will be cancelled.

These are possible cases.

| Case | PylonItem placeable in design | Can trigger BlockPlaceEvent when use |  BlockListener's behavior |
| --- | --- | --- | --- |
| 1 | √ | √ | pass |
| 2 | √ | × | pass |
| 3 | × | √ | cancelled event |
| 4 | × | × | cancelled event |

Totally, BlockListener will handles BlockPlaceEvent incorrectly in case 3. Is there any method to avoid it without change the parameter `event.itemInHand`?

balugaq:
Closes #222 

Also took the chance to learn Kotlin, let me know if there's a better way to structure this :)

Main changes:
1. Fix #222 - See **BlockListener**, **NotPlaceable**
2. Check if the event is a fake event - See **BlockListener**, **PylonUtils**
3. Add comments in **PylonBlock**
4. Add **BlockCreateContext.PluginGenerate**, which is a different case of creating of PylonBlock.

### DIFF
--- a/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/PylonBlock.kt
+++ b/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/PylonBlock.kt
@@ -116,8 +116,6 @@ open class PylonBlock protected constructor(val block: Block) {
     /**
      * Called when the block is saved
      *
-     * As known as `onSaveBlock`
-     *
      * @see serialize
      */
     open fun write(pdc: PersistentDataContainer) {}
--- a/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/PylonBlock.kt
+++ b/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/PylonBlock.kt
@@ -12,7 +12,9 @@ import io.github.pylonmc.pylon.core.config.Config
 import io.github.pylonmc.pylon.core.config.Settings
 import io.github.pylonmc.pylon.core.datatypes.PylonSerializers
 import io.github.pylonmc.pylon.core.event.PylonBlockDeserializeEvent
+import io.github.pylonmc.pylon.core.event.PylonBlockLoadEvent
 import io.github.pylonmc.pylon.core.event.PylonBlockSerializeEvent
+import io.github.pylonmc.pylon.core.event.PylonBlockUnloadEvent
 import io.github.pylonmc.pylon.core.registry.PylonRegistry
 import io.github.pylonmc.pylon.core.util.position.BlockPosition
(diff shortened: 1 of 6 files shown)


pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/util/PylonUtils.kt
```diff
@@ -171,3 +172,7 @@ fun Component.withArguments(args: List<TranslationArgumentLike>): Component {
     }
     return result.children(result.children().map { it.withArguments(args) })
 }
+
+fun isFakeEvent(event: Event): Boolean {
+    return event.javaClass.name.contains("Fake");
```
OhmV-IR, line 177: This is okay, but I think using an annotation class would be more clear & reliable.

OhmV-IR (commented):
Not sure what these "fake" events are and where the classes for them are, can't find any in papermc and pretty sure there aren't any in pylon.

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/util/PylonUtils.kt
balugaq: Not only for Pylon, but also for other plugins

Case 1
https://github.com/mcMMO-Dev/mcMMO/tree/master/src/main/java/com/gmail/nossr50/events/fake

Case 2
https://github.com/Ghost-chu/QuickShop-Reremake/blob/dd52504e3271c91a8b4573c442a7174d2e743f39/src/main/java/org/maxgamer/quickshop/util/PermissionChecker.java#L134

But there's no case 2 check

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/PylonBlock.kt
```diff
@@ -96,6 +115,10 @@ open class PylonBlock protected constructor(val block: Block) {
 
     /**
      * Called when the block is saved
+     *
+     * As known as `onSaveBlock`
```
Seggan, line 119: It's not known as `onSaveBlock`, I would remove thi until we actually get around to renaming it

OhmV-IR (commented):
Looks fine cw

PR #253, opened 2025-08-28
fix runSnapshotServer gradle task

OhmV-IR:
core jar was getting named pylon-core--SNAPSHOT.jar while the runServer action looked for the pylon-core-SNAPSHOT.jar file so it wouldn't copy it and base would fail to load


OhmV-IR:
not the way seggan wanted

PR #254, opened 2025-08-28
Conditional placement

OhmV-IR:
Allows a block to provide a lambda for the conditions in which it should be placed, eg water placer not being placed in the nether


Seggan (changes requested):
See comments on base PR

PR #271, opened 2025-09-17
Fix pylon items not playing nice with vanilla recipe

LordIdra:


LordIdra:
Yoloing into master so we can progress with playtesting, feel free to retroactively review and leave comments here

LordIdra:
based on https://github.com/pylonmc/pylon-core/pull/258/files

PR #287, opened 2025-09-18
Fix #269

### LINKED ISSUE #269: Automatically give recipe to recipe book if no research required
There's no recipe for manual core drill lever in recipe book because it is never unlocked

Seggan:
PRing instead of master push cause the code is more complicated than other fixes


LordIdra (changes requested):
The performance implications of this are quite bad, as we are iterating through every research once for every item in a player's inventory every single time they pick up or click an item. Wouldn't it make more sense here to listen to player join events and discover all the recipes for all items without a research when they join?

Seggan:
No

LordIdra (approved):
Not tested ingame but trust you've tested and it works

Issue #191, opened 2025-07-25
Blockstorage does not handle sensitive blocks


LordIdra:
for fuck's sake

PR #293, opened 2025-09-19
Fix blockstorage not handling sensitive blocks

LordIdra:
closes https://github.com/pylonmc/pylon-core/issues/191

### DIFF
--- a/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/BlockListener.kt
+++ b/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/BlockListener.kt
@@ -130,6 +130,7 @@ internal object BlockListener : Listener {
     private fun blockRemove(event: BlockDestroyEvent) {
         if (BlockStorage.isPylonBlock(event.block)) {
             BlockStorage.breakBlock(event.block, BlockBreakContext.Destroyed(event))
+            event.setWillDrop(false)
         }
     }
 


pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/BlockListener.kt
```diff
@@ -109,26 +107,32 @@ internal object BlockListener : Listener {
         }
     }
 
-    // TODO this might be dropping vanilla blocks in place of Pylon blocks
-    // TODO this will not respect pylon block break events being cancelled
```
LordIdra, line 113: tested to make sure these are no longer needed

OhmV-IR:
Are we sure this is a good idea unless explicitly specified by a block scheme? Eg placing a solar lens wrong and then losing it would be a big pain

LordIdra:
> Are we sure this is a good idea unless explicitly specified by a block scheme? Eg placing a solar lens wrong and then losing it would be a big pain

Uhhh... what?

OhmV-IR (changes requested):
looks fine cw, but has ingame issues.
Repo steps:
1. place hydraulic core drill on ground
2. place core drill lever on the drill
3. break drill
A core drill lever, hydraulic core drill, **and a regular lever** drop
Just placing a vanilla lever on the block then breaking it has correct behavior

LordIdra:
> looks fine cw, but has ingame issues. Repo steps:
> 
> 1. place hydraulic core drill on ground
> 2. place core drill lever on the drill
> 3. break drill
>    A core drill lever, hydraulic core drill, **and a regular lever** drop
>    Just placing a vanilla lever on the block then breaking it has correct behavior

Thanks for the catch. Not sure why I didn't catch this when testing. Fixed now

Intybyte (approved):

Intybyte:
Maybe try with BlockPhysicsEvent?

LordIdra:
> Maybe try with BlockPhysicsEvent?

Why..?

PR #304, opened 2025-09-21
[Feature] Armor Textures

### LINKED ISSUE #205: Extra component for armor layers
I'd like to suggest adding an extra component (nbt tag) to armor pieces to be able to implement armor layer texture.
(without need of extra mod) 

Image

Image

basically the `equippable` component is used to define where the item can be equipped with the following `slot: <head,chest,legs,etc>`
the main part is `assets_id: <id>` the ID can be whatever you want, I'd suggest <armor_name>_armor to keep it simple

for example
assets_id: ferroduralum_armor
now
since the item has this component
I can create for my resourcepack assets/minecraft/equipment/ferroduralum_armor.json and the rest is up to me 

Image

JustAHuman-xD:
(this relies on #303 and is currently based on merging into it so that reviewing is easier, once that PR is merged I'll change the target branch to master)

Resolves #205 

Compared to Block Textures armor textures are incredibly simple, however, slightly less nice. Minecraft has no sort of fallback system for equipment definitions, so if we set the equippable components asset id to say "pylonbase:bronze", if the player doesn't have a resource pack, the armor will just not render at all. (it *doesn't* show missing textures though). Because of that, custom armor textures have to be disabled by default, and opt-in.

Implementation wise, there is now a PylonArmor interface for items, which just has an abstract field for the assetId, and a ArmorTextureEngine. Whenever a player has customArmorTextures enabled, we will intercept the packets to do with entity equipment & the players inventory, and will change the equippable component for the client (assuming it hasn't been changed by something else). There is also handling for creative inventory packets, as when a player is in creative, they have full control over their inventory and we don't want it to change the *actual* equippable component.

Right now there are only 2 real "issues" (afaik)
1. If an entity equips custom armor while in tracked by a player, when the player changes their customArmorTexture option, the entitys armor will not update until the entire entity is resent or its equipment gets updated. (This doesn't really matter though)
2. When the player is wearing custom armor and changes the option, their armor wont update until something updates their inventory (i.e. picking up an item, moving an item in the inventory, etc) which also doesn't really matter. I tried to use Player#updateInventory to make it just send updates, but the method just... didn't do anything lol


pylon-core/src/main/resources/lang/en.yml
```diff
@@ -167,6 +167,9 @@ guide:
       high:
         name: "<gray>Culling Preset: <red>High <dark_gray>(Click to cycle)"
         lore: "<arrow:#3a2939> Aggressive culling always showing entities within %alwaysShowRadius% blocks and heavily culling entities within %cullRadius% blocks (only culling if there are %maxOccludingCount% occluding blocks between your eyes and the block)"
+    toggle-armor-textures:
+      name: "Toggle Custom Armor Textures"
+      lore: "<arrow:#3a2939> If you have a resource-pack that adds <bold>worn <gray>(not item) pylon armor textures, enable this to use them, otherwise leave it disabled"
```
Seggan, line 172:
```suggestion
      lore: "<arrow:#3a2939> If you have a resource pack that adds <bold>worn <gray>(not item) Pylon armor textures, enable this to use them, otherwise leave it disabled"
```

Seggan (approved):
LGTM otherwise, untested because idk how lol

Seggan:
Nah I'm good, I hate testing :P

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/resourcepack/armor/ArmorTextureEngine.kt
```diff
@@ -0,0 +1,133 @@
...
+        packet.equipment.forEach { equipment ->
+            equipment.item = addArmorTexture(equipment.item)
+        }
+    }
+
+    private fun addArmorTexture(packetStack: ItemStack?) : ItemStack? {
+        if (packetStack == null) return null
+        val stack = SpigotConversionUtil.toBukkitItemStack(packetStack)
+        val item = PylonItem.fromStack(stack)
+        if (item != null && item is PylonArmor) {
```
OhmV-IR, line 91:
```suggestion
        if (item is PylonArmor) {
```
is check will be false if item is null, so the explicit item != null check is not needed

OhmV-IR (commented):
just cw comment, haven't tested ig yet

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/item/base/PylonArmor.kt
```diff
@@ -0,0 +1,7 @@
+package io.github.pylonmc.pylon.core.item.base
+
+import net.kyori.adventure.key.Key
+
+interface PylonArmor {
+    val equipmentType: Key
```
LordIdra, line 6: Could we get a javadoc here to say what kind of key this should be? (It is also probably worth documenting that the interface just allows the armour to be given a texture)

LordIdra (approved):
Only comment is docs. Need documenting overall - ideally javadocs or high level documentation - but happy to leave this till later as long as it's done

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/resourcepack/armor/ArmorTextureEngine.kt
```diff
@@ -0,0 +1,133 @@
...
+import io.papermc.paper.datacomponent.item.Equippable
+import org.bukkit.entity.Player
+import org.bukkit.persistence.PersistentDataType
+
+object ArmorTextureEngine : PacketListener {
+    val customArmorTexturesKey = pylonKey("custom_armor_textures")
+    val armorSlots = setOf(5, 6, 7, 8)
+
+    @get:JvmStatic
+    @set:JvmStatic
```
Seggan, line 29:
```suggestion
    @JvmStatic
```

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/resourcepack/armor/ArmorTextureEngine.kt
```diff
@@ -0,0 +1,133 @@
...
+    var Player.customArmorTextures: Boolean
```
Seggan, line 30:
```suggestion
    var Player.customArmorTexturesEnabled: Boolean
```

Seggan (approved):
Some small bits I found while re-reviewing

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/resourcepack/armor/ArmorTextureEngine.kt
JustAHuman-xD: Why?

JustAHuman-xD: the fact that it returns a boolean I think makes it clear but I can add enabled if its *really* necessary, its just already a long name

Seggan:
Woops hit wrong button

PR #305, opened 2025-09-21
Fix block break events always being cancelled when blocks are pylon blocks

OhmV-IR:
No real way to test, but talked through the logic with justin and idra in vc

### DIFF
--- a/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/BlockStorage.kt
+++ b/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/BlockStorage.kt
@@ -245,14 +245,14 @@ object BlockStorage : Listener {
     fun breakBlock(
         blockPosition: BlockPosition,
         context: BlockBreakContext = BlockBreakContext.PluginBreak()
-    ) {
+    ) : MutableList<ItemStack>? {
         require(blockPosition.chunk.isLoaded) { "You can only break Pylon blocks in loaded chunks" }
 
-        val block = get(blockPosition) ?: return
+        val block = get(blockPosition) ?: return null
 
         val event = PrePylonBlockBreakEvent(blockPosition.block, block, context)
         event.callEvent()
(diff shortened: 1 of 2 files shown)


pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/BlockStorage.kt
```diff
@@ -245,14 +245,14 @@ object BlockStorage : Listener {
     fun breakBlock(
         blockPosition: BlockPosition,
         context: BlockBreakContext = BlockBreakContext.PluginBreak()
-    ) {
+    ) : MutableList<ItemStack>? {
```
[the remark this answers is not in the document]
LordIdra: yeah this should be immutable because the drops have already been droopped

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/BlockStorage.kt
```diff
@@ -245,14 +246,14 @@ object BlockStorage : Listener {
     fun breakBlock(
         blockPosition: BlockPosition,
         context: BlockBreakContext = BlockBreakContext.PluginBreak()
-    ) {
+    ) : List<ItemStack>? {
```
Seggan, line 249: Needs docs on when it might return null vs an empty list

Seggan (changes requested):

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/BlockStorage.kt
JustAHuman-xD: there already is

Seggan: Oops

Seggan (approved):

PR #311, opened 2025-09-21
Switch ChunkPosition and BlockPosition to use UUID

### LINKED ISSUE #294: Change ChunkPosition to store the World UUID instead of (weak) references
That way it can handle loading & unloading of worlds

JustAHuman-xD:
Resolves #294


Seggan (approved):

LordIdra:
Will break every block on the test server. Do not merge yet. Marking as draft so we don't accidentally merge this

JustAHuman-xD:
> Will break every block on the test server. Do not merge yet. Marking as draft so we don't accidentally merge this. (nvm can't but still, don't merge)

wait what breaks?

PR #313, opened 2025-09-21
[Feature] Pylon Guide on First Join

### LINKED ISSUE #306: Give player Pylon guide when they first join
!!!!!!

(toggle option in config, SET TO TRUE BY DEFAULT)

JustAHuman-xD:
resolves #306

### DIFF
--- a/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/config/PylonConfig.kt
+++ b/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/config/PylonConfig.kt
@@ -8,7 +8,7 @@ object PylonConfig {
     private val config = Config(PylonCore, "config.yml")
 
     @JvmStatic
-    val firstJoinPylonGuide = config.getOrThrow("pylon-guide-on-first-join", ConfigAdapter.BOOLEAN)
+    val pylonGuideOnFirstJoin = config.getOrThrow("pylon-guide-on-first-join", ConfigAdapter.BOOLEAN)
 
     @JvmStatic
     val defaultTickInterval = config.getOrThrow("default-tick-interval", ConfigAdapter.INT)
--- a/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/config/PylonConfig.kt
+++ b/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/config/PylonConfig.kt
@@ -8,6 +8,9 @@ object PylonConfig {
     private val config = Config(PylonCore, "config.yml")
 
     @JvmStatic
+    val firstJoinPylonGuide = config.getOrThrow("pylon-guide-on-first-join", ConfigAdapter.BOOLEAN)
+
+    @JvmStatic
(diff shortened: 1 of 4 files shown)


Seggan (commented):
Why do you use two events? Why not simplify it and just use PlayerJoinEvent?

OhmV-IR:
> Why do you use two events? Why not simplify it and just use PlayerJoinEvent?

+1

JustAHuman-xD:
> Why do you use two events? Why not simplify it and just use PlayerJoinEvent?

Namely because I assumed that by the time of the PlayerJoinEvent the player was considered to have played before, but I can check

JustAHuman-xD:
You guys were correct, at least for a setup with only these plugins, I'm not fully convinced there isn't a case where it fails, but itd require another plugin with the lowest priority on player join, that runs before ours, to tell the server to save the players data. If that happens this will fail as hasPlayedBefore checks if loading the player nbt file is null or not

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/content/guide/PylonGuide.kt
```diff
@@ -79,6 +86,13 @@ class PylonGuide(stack: ItemStack) : PylonItem(stack), PylonInteractor {
         @JvmStatic
         val settingsAndInfoPage = SettingsAndInfoPage()
 
+        @EventHandler(priority = EventPriority.LOWEST)
+        fun join(event: PlayerJoinEvent) {
```
Seggan, line 90:
```suggestion
        private fun join(event: PlayerJoinEvent) {
```

Seggan (approved):
LGTM otherwise

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/config/PylonConfig.kt
```diff
@@ -7,6 +7,9 @@ object PylonConfig {
 
     private val config = Config(PylonCore, "config.yml")
 
+    @JvmStatic
+    val firstJoinPylonGuide = config.getOrThrow("pylon-guide-on-first-join", ConfigAdapter.BOOLEAN)
```
LordIdra, line 11: nit but config val should be named the same as the config key (pylonGuideOnFirstJoin). (also `give-pylon-guide-on-first-join`?)

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/content/guide/PylonGuide.kt
```diff
@@ -79,6 +86,13 @@ class PylonGuide(stack: ItemStack) : PylonItem(stack), PylonInteractor {
         @JvmStatic
         val settingsAndInfoPage = SettingsAndInfoPage()
 
+        @EventHandler(priority = EventPriority.LOWEST)
```
LordIdra, line 89: Should be documented why this is LOWEST probably

LordIdra (changes requested):

LordIdra (approved):

PR #323, opened 2025-09-24
Better handling of pylon block explosions

### LINKED ISSUE #301: Exploded Pylon Blocks NEVER Drop
Pylon blocks should either always drop to explosions, or should work like vanilla blocks and have a chance to drop/not drop

Intybyte:
Closes #301


JustAHuman-xD:
Maybe consider basing this off of https://github.com/pylonmc/pylon-core/pull/305 as it makes large changes to this are as well

Intybyte:
i forgot that in the event the block is also passed for some reason and added unnecessary stuff also, maybe it is better that i nuke this and just add stuff to the other PR

PR #324, opened 2025-09-24
Update unmodifed files

### LINKED ISSUE #261: Language and settings files not updated in production
How the hell did it take us this long to realise?

If we change a lang file, the changes will not be propagated to any production servers. So if the lore of an item is updated, that will **never** be propagated to servers until they reset (or until manual intervention). (The same goes for settings, recipes, config, etc)

Right now we plan to solve for recipes this with an 'unlock' command. This isn't really practical to do for everything.

It's also not practical to provide migration scripts for configs like other plugins do.

Proposed solution:
- When Pylon starts, store the hash of every file Pylon uses to persistent storage. If a new file is registered later, hash it and store the hash when it's first used
- When Pylon or an addon starts:
  - Check the hash of each file **in the resource** and compare it to the corresponding stored hash
  - If they are different, that file needs to update, check the stored hash and compare it to the current hash
  - If that is the same, the file has not been modified since last update, so just replace the file with the new version
  - If that is not the same, the file has been modified. Back up the old file, replace the file with the new one, and then log a big message in console and potentially also log a message that must be dismissed when an administrator logs on.
  - Either way, updated the stored hash to the hash of the new file.
- When Pylon starts again, recompute the hashes and compare them to the ones in persistent storage. 

Yeah it's complicated, but can be neatly abstracted and isolated, and solves the problem effectively. But if anyone has any other ideas curious to hear them.

Major blocker for production release so marking as high priority.

Intybyte:
Closes #261

### DIFF
--- a/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/addon/PylonAddon.kt
+++ b/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/addon/PylonAddon.kt
@@ -4,6 +4,7 @@ import io.github.pylonmc.pylon.core.PylonCore
 import io.github.pylonmc.pylon.core.block.BlockStorage
 import io.github.pylonmc.pylon.core.config.Config
 import io.github.pylonmc.pylon.core.config.ConfigSection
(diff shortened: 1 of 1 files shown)


Seggan:
I concur

Issue #230, opened 2025-08-16
Symbol-Based Precision Search


balugaq:
In multi-mod environments, players need efficient item search. JEI's `@` symbol (e.g., `@enderio` for EnderIO items) and my SF addon JustEnoughGuide's similar implementation have proven effective for targeted searches.

I propose expanding this concept for Pylon with multiple search symbols:

| Symbol | Function | Example |
|--------|----------|---------|
| `@`    | Items from specific mod/plugin | `@base` |
| `!`    | Item name contains keyword | `!water` |
| `~`    | Item material contains keyword | `~stone` |
| `#`    | Recipe includes ingredient with keyword | `#chain` |
| `$`    | Items from specific recipe type | `$furnace` |

Symbols can be combined for precise filtering:
- `@enderio !pipe` – EnderIO items with "pipe" in name
- `~wood #iron` – Wooden items using iron in recipes
- `$crafting @base !chest` – PylonBase chests made via crafting table

PR #329, opened 2025-09-25
Search Improvements

### LINKED ISSUE #243: Text in search box not saved when going back

JustAHuman-xD:
Resolves #243 #230 #290 

Also adds/changes a few ui related things I found while working on it.


pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/guide/pages/base/SearchPage.kt
```diff
@@ -53,27 +77,89 @@ abstract class SearchPage (key: NamespacedKey, material: Material) : SimpleStati
...
 
-    open fun getItems(player: Player, search: String): List<Item> = getItemNamePairs(player, search).map {
-        val name = it.second
-        val distance = searchAlgorithm.distance(name, search)
-        val weight = when {
-            name == search -> 0.0
-            name.startsWith(search) -> 1.0 + distance
-            name.contains(search) -> 2.0 + distance
-            else -> 3.0 + distance
+    open fun getItems(player: Player, search: String): List<Item> {
```
LordIdra, line 80: This method is quite hard to read now - do you think we could split it up and/or comment it a bit? Maybe the blocks inside when could be abstracted. It seems like we could possibly turn the search terms into their own classes and implement the methods there, which would be a nice separation of concerns.

Also, is there any reason that a normal word (eg 'bronze') is not treated as a search term? It seems like it would be cleaner to do that but maybe there's a reason you haven't?

LordIdra (changes requested):
Nice work

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/guide/pages/base/SearchPage.kt
LordIdra: Gotcha. When I read 'term' I generally think of one word in the query - like `(bronze) (@pylonbase) ($bruh)`. Do you think it would be worth slightly changing the logic to group terms like this instead? My thinking is that then you could unify all the logic under the 'term' idea and process each word independently.

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/util/PylonUtils.kt
```diff
@@ -339,3 +342,9 @@ fun <T> persistentData(
 
 val Block.replaceableOrAir: Boolean
     get() = type.isAir || isReplaceable
+
+fun Component.render(locale: Locale = Locale.ROOT): Component
```
Seggan, line 346:
```suggestion
fun Component.translate(locale: Locale = Locale.ROOT): Component
```

Seggan (commented):
LGTM CW, untested

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/util/PylonUtils.kt
JustAHuman-xD: thats a good point that hits both the names render and translate, I can add a doc for it

Seggan: I would still prefer something to indicate what it does without looking at docs, especially since its an extension method

JustAHuman-xD: and if someone doesn't understand it they can always just control click into it and immediately see what it does, or just hover over it and see what it does (assuming they have documentation included in the project)

LordIdra: I mean I'd prefer to just use GlobalTranslator.render instead of adding another level of indirection as it seems a bit of a pointless abstraction, just style/preference. (there's your 3rd opinion). Whether it's render or translation or globaltranslate, I don't really care, I'd be confused because I hadn't seen the method before and would have to check the docs either way, flip a coin

JustAHuman-xD: I can just remove the extension if its this much of a problem, it was just meant to be a nice short hand

PR #333, opened 2025-09-26
Fixup Dokka

Seggan:
Because Dokka is stupid


pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/util/PylonUtils.kt
```diff
@@ -340,6 +340,9 @@ fun <T> persistentData(
     default: T
 ) = persistentData(key, type) { default }
 
+val Block.replaceableOrAir: Boolean
```
LordIdra, line 343: What's this? Doesn't seem like it's part of this PR - mistake when rebasing maybe?

LordIdra:
Please could you give more info on how to run this. I have run dokkaGenerate and I am still seeing the issues that I think this pr is supposed to fix

image
image

LordIdra:
Wait sorry I didn't see your docs PR. Will have a look through those

Seggan:
https://pylonmc.github.io/contributing/getting-started/#custom-dokka

LordIdra:
image
Still seeing companions show up. Is that supposed to be fixed?

LordIdra:
> Most of those instances is because the companion implements listener, so it's "potentially useful" as far as dokka knows

Oh bruh. Should we open another issue for that?

Seggan:
No because that's completely logical behavior. Java *might* want to pass the companion as a listener somewhere. Of course, it won't because that's not how listeners work, but Dokka doesn't know that. While I can imagine suggesting a "ignore these interfaces" option or something I can't imagine it will be accepted due to it being a super niche thing. I doubt we could use a Dokka plugin for this due to [insert dokka internal mumbo jumbo I dont care to explain right now]

LordIdra:
> No because that's completely logical behavior. Java _might_ want to pass the companion as a listener somewhere. Of course, it won't because that's not how listeners work, but Dokka doesn't know that. While I can imagine suggesting a "ignore these interfaces" option or something I can't imagine it will be accepted due to it being a super niche thing. I doubt we could use a Dokka plugin for this due to [insert dokka internal mumbo jumbo I dont care to explain right now]

ugh that's really annoying. ffs. Ok I am happy to merge then. Thanks for fixing this dumpster fire lol

LordIdra (approved):

PR #363, opened 2025-10-03
Add automatic handling for PylonItems

Intybyte:
Made to fix https://github.com/pylonmc/pylon-base/issues/281, I hope there is a better way to do it, Seggan worked on this the most so he probably has a better solution


Seggan:
Already fixed (in a better way) by #351

PR #376, opened 2025-10-04
Fix build

LordIdra:
Companion to https://github.com/pylonmc/pylon-base/pull/319

Fixes screw up during merge that meant PylonFluidPointDirectional failed to compile.

Removes display recipe type because it's no longer needed (and tbh we should strongly discourage addons using display recipes anyway - it made sense in the past but with recipe configs and some of the more recent changes I don't think it's needed any more).


JustAHuman-xD (approved):

OhmV-IR (approved):
lgtm cw & ig

PR #352, opened 2025-09-30
Item Builder Update

JustAHuman-xD:
Supercedes #325 
Based on #341 as it modified the stack builder already and started on some of the changes I needed for this

Adds a PylonItemStackBuilder with new utils and moves the previous pylon item related ones from ItemStackBuilder to it.

Adds a number of new utility methods for both.

Gives ui items specific keys to be identifiable, also gives PylonFluid items a specific key and adds a method to get the fluid from the item stack representation

PR #377, opened 2025-10-05
MultiblockRefreshEvent, id provider fix

JustAHuman-xD:
Relies on #352 

Adds the multiblock refresh event which is needed to finish https://github.com/pylonmc/pylon-base/pull/255
Also fixes an issue with the packet based entities, turns out the id provider gets overwritten by the init method for some reason, so had to move its assignment to after initializing, also stopped tracking platform entities as we don't need to.


JustAHuman-xD:
> What does refresh mean for multiblocks?

To quote the existing kdoc

> Called when the multiblock is refreshed (i.e., was formed before, and still is).
     This happens when something causes the Multiblock to re-check if it is formed,
     and it still is.

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/util/Octree.kt
```diff
@@ -127,9 +127,39 @@ open class Octree<N>(
...
 
+    override fun iterator(): Iterator<N> = OctreeIterator(this)
+
     companion object {
         // LISTEN IF THESE ARE INSANE FIX THEM, THIS ISN'T MY STRONG SUIT - JustAHuman
         private const val DEFAULT_MAX_DEPTH = 2048
         private const val DEFAULT_MAX_ENTRIES = 128
     }
+
+    private class OctreeIterator<N>(root: Octree<N>) : Iterator<N> {
```
Seggan, line 138: Can't this be an inner class instead of having `root` in the constructor?

JustAHuman-xD: I mean, I guess? I prefer it not to be, but I can change it I suppose, I don't really have a reason for not wanting it other than preference lol.

Seggan: Yeah I understand, was just wondering

Issue #327, opened 2025-09-25
Allow hiding buttons in settings


LordIdra:
Needed to remove texture buttons if the server has texture packs disabled.

Right now you can use if/else statements but it'll leave gaps in the settings.

To fix this we probably just want to make the settings GUI a paged GUI. But we might want to change this in future.

LordIdra:
resource pack settings will have their own page soon but we;ll want this regardless

PR #386, opened 2025-10-06
Settings system & Waila update

### LINKED ISSUE #227: Action bar WAILA
also add a guide setting, players can choose displaying WAILA at bossbar or action bar.

JustAHuman-xD:
Requires #377 
Resolves #227 #327 #398 #397 

Changes the SettingsPage to have no buttons by default but buttons can be added with SettingsPage#addSetting, adds a waila settings button (which expands into the different options), a resource pack settings option (which holds block texture & armor texture settings), and added a setting for disabling research effects for players.

There are now a few default settings buttons that can be used for simple options (toggles, cycles, and numeric settings), the previously hardcoded toggle & cycle buttons have been removed and have had their usages replaced with these new buttons.

The waila can now be globally disabled by setting the tick interval to be <= 0. I reworked the waila itself a little bit to allow for ActionBar waila and for more configurability, as well as internally changing it to use only 1 raycast call, added events for both block and entity wailas, and added entity waila overrides, as only blocks could have their waila overriden. I also moved the waila out of the blocks package, as it isn't only for blocks.

I remember seeing somewhere a discussion on allowing players to enable the waila for vanilla blocks/entities but I couldn't find an issue for it, but it is implemented by this as well.

While adapting the system I also made it so servers can disable armor textures, and added the option to force block or armor textures as well. (For servers that have a server-side resource pack for ex)

### DIFF
--- a/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/PylonCore.kt
+++ b/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/PylonCore.kt
@@ -22,8 +22,8 @@ import io.github.pylonmc.pylon.core.entity.EntityStorage
 import io.github.pylonmc.pylon.core.entity.PylonEntity
 import io.github.pylonmc.pylon.core.fluid.connecting.ConnectingService
 import io.github.pylonmc.pylon.core.guide.button.PageButton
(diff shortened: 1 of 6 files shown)


Seggan (changes requested):
~~Any WAILA problems can now be `git blame`d on Justin~~

PR #438, opened 2025-10-21
Fix pistons and add StructureGrowEvent support

### LINKED ISSUE #437: Multiblock don't detect structure grows and piston handling doesn't behave as expected
The pit kiln does not start (according to WAILA) when the last "placed" block of the multiblock is dirt changing into podzol from a 2x2 spruce growing nearby.
It runs normally when updated by breaking and replacing one of its blocks.

Intybyte:
Fixes #437


haseir:
Don't you also need to check other events (e.g. BlockSpreadEvent) for moss or sculk spread, enderman pickup etc.?

balugaq (approved):
Tested and works well

LordIdra (approved):

PR #441, opened 2025-10-24
Don't warn for missing researches

### LINKED ISSUE #408: Remove warning about missing `researches.yml`
If it's missing its probably intentional

Intybyte:
Fixes #408


LordIdra (changes requested):
Small thing but missing docs. Just leaving these comments so we get into the habit of writing docs for stuff like this. A sentence about warnMissing will suffice

LordIdra (approved):

PR #476, opened 2025-11-02
Add open pylon guide command(/py open_guide)

condblock:


condblock:
> Currently, the behaviour is for `/py` to open the guide. This may be unintuitive for some Slimefun users, but I think it's the decision in the long run since it's so convenient. I'd prefer to avoid adding a `open_guide` command to duplicate this functionality tbh

Oh, I didn't know the /py command open the guide before.

PR #482, opened 2025-11-14
Add PylonItem#fromStack(ItemClass, ItemStack)

Intybyte:
Fixes https://github.com/pylonmc/pylon-core/issues/459


Seggan (approved):
You might wanna add a reified one too

PR #496, opened 2025-11-21
Add PylonItemSchema#fromStack

Intybyte:
Closes https://github.com/pylonmc/pylon-core/issues/491


pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/item/PylonItemSchema.kt
```diff
@@ -71,5 +72,11 @@ class PylonItemSchema @JvmOverloads internal constructor(
 
     companion object {
         val pylonItemKeyKey = pylonKey("pylon_item_key")
+
+        @JvmStatic
+        @Contract("null -> null")
+        fun fromStack(stack: ItemStack) : PylonItemSchema? {
+            return PylonItem.fromStack(stack)?.schema
```
LordIdra: Performance - you should just read the PDC key here and index into the pylon item registry to avoid instantiating a PylonItem. I envision us using this in a lot of places in the future for performance's sake so this would be good to do now

LordIdra (approved):

OhmV-IR (approved):

Issue #488, opened 2025-11-17
Pylon items can be used in Cartography Table


JustAHuman-xD:
They can be used as their vanilla counterparts

PR #497, opened 2025-11-21
CartographyTable fix

Intybyte:
Fixes https://github.com/pylonmc/pylon-core/issues/488

Comment needs to be addressed

### DIFF
--- a/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/recipe/RecipeListener.kt
+++ b/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/recipe/RecipeListener.kt
@@ -88,7 +88,6 @@ internal object PylonRecipeListener : Listener {
         val inventory = e.inventory
         val hasPylonItems = inventory.any { it.isPylonAndIsNot<VanillaCraftingItem>() }
 
-        // Should we add a RECIPE_TYPE For this or just let addon devs handle it?
         if (hasPylonItems) {
             inventory.result = null
         }


pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/recipe/RecipeListener.kt
```diff
@@ -82,6 +83,17 @@ internal object PylonRecipeListener : Listener {
         }
     }
 
+    @EventHandler(priority = EventPriority.LOWEST)
+    private fun onCartography(e: CartographyItemEvent) {
+        val inventory = e.inventory
+        val hasPylonItems = inventory.any { it.isPylonAndIsNot<VanillaCraftingItem>() }
+
+        // Should we add a RECIPE_TYPE For this or just let addon devs handle it?
```
Seggan: @JustAHuman-xD i believe has something to say on this (I forgot exactly what)

JustAHuman-xD: Definitely an addon, the cartography table doesn’t allow putting in 99% of items by default and requires slightly cursed nms shenanigans to do so, which should NOT be in core

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/recipe/RecipeListener.kt
```diff
@@ -82,6 +83,16 @@ internal object PylonRecipeListener : Listener {
         }
     }
 
+    @EventHandler(priority = EventPriority.LOWEST)
+    private fun onCartography(e: CartographyItemEvent) {
+        val inventory = e.inventory
+        val hasPylonItems = inventory.any { it.isPylonAndIsNot<VanillaCraftingItem>() }
```
LordIdra: Could have a `VanillaCartographyItem` but I think this is fine, and niche enough that it wouldn't be used

PR #499, opened 2025-11-22
Fix recipe autocompletion

Intybyte:
How am feeling rn https://www.youtube.com/shorts/sURzzeCHUUc

Fixes https://github.com/pylonmc/pylon-core/issues/475

Requires massive testing, has some dead code that I will remove once it is confirmed this abomination works as expected

Client still thinks some recipes are available however and I think that is a client side thing so we can't fix it


Seggan (changes requested):
ngl I'm very skeptical of this code. A lot of it seems to be copied from Paper/NMS, which has many problems, most importantly licensing. Also, this is frankly a huge amount of code touching NMS (700 lines :monkaS:), and IMO it might just not be worth it to have to debug this code every time we update.

Seggan:
NMS is *not* a gray area, there's a reason why Bukkit was taken down and why Spigot and Paper only open source patches. Licensing copied code is *absolutely* a concern.

Intybyte:
> NMS is _not_ a gray area, there's a reason why Bukkit was taken down and why Spigot and Paper only open source patches. Licensing copied code is _absolutely_ a concern.

No that was not the case,  bukkit went down due to the fact that it was an open source project that was secretly owned by mojang, and once the news hit the public a major contributor DMCAd the whole thing down, also if we take down every project that uses nms then we gonna take down all mods

Yeah ik some of the code is copied, and honeslty we can fix that part somehow

Seggan (approved):
Woops wrong button

Intybyte:
> @Intybyte I think we are all in agreement(ish) to merge this. could we get an explanation somewhere in this PR of what the hell is going on for future maintainers? after that I am happy to merge

Already did, is there some sections where you would like more documentation or more accurate one?

PR #512, opened 2025-11-26
Expose blockClass in PylonBlockSchema

Intybyte:


JustAHuman-xD:
Can you also change the access of the item class so that the method name on java isn't "getItemClass$pylon_core" and is just getItemClass

Issue #505, opened 2025-11-23
Pipe Dupe


balugaq:
https://github.com/user-attachments/assets/2c93dcdb-f399-40d1-a2a2-2a30e49b6fcb

video ^
steps:
1. hold A pipe on main hand and B pipe on offhand
2. right click the ground
3. swap hands by pressing down F
4. right click again

result: placed A pipe but consumed B pipe

expected:
cancel pipe pre-place when swapping hands

LordIdra:
Oh good find. Really easy fix - just need to listen for off hand swap event and cancel if player is currently placing a pipe.

(Might also be worth checking if swapping the pipe in your inventory instead does the same thing. I suspect it might)

PR #518, opened 2025-12-10
Fix pipe TRANSMUTATION

Intybyte:
Fixes https://github.com/pylonmc/pylon-core/issues/505


Seggan (approved):
You don't need the `!= null` because `is` already checks for that but otherwise LGTM

LordIdra (approved):
Tested

PR #520, opened 2025-12-11
Add support for falling blocks

Intybyte:


pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/base/PylonFallingBlock.kt
```diff
@@ -0,0 +1,68 @@
...
+import org.bukkit.block.Block
+import org.bukkit.entity.FallingBlock
+import org.bukkit.event.entity.EntityChangeBlockEvent
+import org.bukkit.persistence.PersistentDataContainer
+import org.bukkit.persistence.PersistentDataType
+
+interface PylonFallingBlock {
+    fun onFallStart(event: EntityChangeBlockEvent) {
+
+    }
```
JustAHuman-xD: the serialized block pdc should be passed to this method

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/base/PylonFallingBlock.kt
```diff
@@ -0,0 +1,68 @@
...
+
+    fun onFallStop(event: EntityChangeBlockEvent, entity: FallingBlockEntity) {
+
+    }
+
+
+    class FallingBlockEntity : PylonEntity<FallingBlock> {
```
JustAHuman-xD: Should have a less generic name, this is the same name as the nms entity

Intybyte: suggestions?

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/BlockStorage.kt
```diff
@@ -99,6 +99,15 @@ object BlockStorage : Listener {
...
     }
 
+    /**
+     * Raw access to the block map, only use this if you know what you are doing and need
+     * access for the internals, if you are trying to place a block use [placeBlock]
+     */
+    @JvmStatic
+    fun rawPlace(pyBlock: PylonBlock) = lockBlockWrite {
+        blocks[pyBlock.block.position] = pyBlock
+    }
```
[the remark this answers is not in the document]
JustAHuman-xD: Or you have to use something else for falling blocks

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/BlockStorage.kt
```diff
@@ -99,6 +101,15 @@ object BlockStorage : Listener {
...
```
LordIdra: This function could break blockstorage in a few ways. See placeBlock for the other things you need to do. It should all be inside the lockBlockWrite block inside placeBlock.

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/BlockStorage.kt
```diff
@@ -261,6 +272,48 @@ object BlockStorage : Listener {
...
 
+    /**
+     * Loads Pylon block data at a location. Only call on the main thread.
+     *
+     * @return The block that was placed, or null if the block placement was cancelled
+     *
+     * @throws IllegalArgumentException if the chunk of the given [blockPosition] is not
+     * loaded, the block already contains a Pylon block, or the block type given by
+     * [key] does not exist.
+     */
```
LordIdra: `@return` and `@throws` have not been updated

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/PylonBlock.kt
```diff
@@ -306,8 +306,8 @@ open class PylonBlock internal constructor(val block: Block) {
         @JvmStatic
         val pylonBlockTextureEntityKey = pylonKey("pylon_block_texture_entity")
 
-        private val pylonBlockKeyKey = pylonKey("pylon_block_key")
-        private val pylonBlockPositionKey = pylonKey("position")
+        val pylonBlockKeyKey = pylonKey("pylon_block_key")
+        val pylonBlockPositionKey = pylonKey("position")
```
LordIdra: Missing `@JvmStatic`

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/base/PylonFallingBlock.kt
LordIdra: Use `PylonSerializers` in this class instead of directly referencing the datatype or using PersistentDataType

LordIdra (changes requested):
Nice work, generally happy with how this is implemented

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/BlockStorage.kt
LordIdra: What do you mean by separate load and place

Intybyte: by giving a CAUSE enum to the event people listening to the event can know which one of the two is taking place

LordIdra: The PrePylonBlockPlaceEvent already has a context..? You can just add a new context. Unless I'm missing something/

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/BlockListener.kt
```diff
@@ -92,6 +94,66 @@ internal object BlockListener : Listener {
...
+        val blockSchema = fallMap[entity.uniqueId] ?: return
+        val relativeItem = PylonRegistry.ITEMS[blockSchema.key]
+        fallMap.remove(entity.uniqueId)
+        if (relativeItem == null) {
+            event.isCancelled = true
+            return
+        }
+
+        event.itemDrop.itemStack = relativeItem.getItemStack()
+    }
```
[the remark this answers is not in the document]
Intybyte: > This and the whole fallMap system I think should be replaced. Not all falling blocks always drop an item (like anvils for example) or they change state when landing & dropping items (again anvils)

There is little i can do about that, there is no way to know if a falling block  will drop items or change state, and i won't make a map for each specific case, unless you have a better idea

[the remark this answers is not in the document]
JustAHuman-xD: I think I misunderstood fall map and you misunderstood my comment. The onDropItem was all I really needed I just thought we could get rid of fallMap but now assuming I've understood you need it because the entity drops the item after it's been removed? So entity storage doesn't track it anymore even though it's still being used in code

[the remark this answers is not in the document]
Intybyte: Yes exactly

PR #521, opened 2025-12-12
Dynamic tick interval

Intybyte:


OhmV-IR (changes requested):
don't we risk a race condition here if you are writing to the pylonBlock.tickInterval.ticks value on the main thread while the async thread is reading it to delay the thread? I think you need a mutex setup for read/writes to the tickInterval value

Intybyte:
> don't we risk a race condition here if you are writing to the pylonBlock.tickInterval.ticks value on the main thread while the async thread is reading it to delay the thread? I think you need a mutex setup for read/writes to the tickInterval value

Yes, but if you know your code is running async to begin with you should already know how to handle such situations

LordIdra:
There was no debate on the impact of an if statement. Are you referring to the tick pausing debate? I was concerned with the impact of a map lookup every tick there, not the impact of an if statement lmao. I was under the impression that atomics use compare and set under the hood so the performance impact is negligible, am I wrong on that?

Seggan:
> There was no debate on the impact of an if statement

https://discord.com/channels/1329177304857055273/1329177972631928915/1449122587694202951

> I was under the impression that atomics use compare and set under the hood so the performance impact is negligible, am I wrong on that?

Atomic CAS is about 10x slower than a nonatomic CAS due to it requiring taking over cache and possibly invalidating it, and in a contested environment a CAS spinloop is even worse (not saying we *shouldn't* do this, I was just half-sarcastically drawing comparison to a recent convo lol)

LordIdra:
> https://discord.com/channels/1329177304857055273/1329177972631928915/1449122587694202951

I get that this is a small thing, but I'd ask that you please don't mischaracterise my words like saying 'debate on the impact of if statements'. It was about the _map lookup_ and not about the if statement itself (as is easily evident in all the context surrounding the message) and I think you can appreciate it's a bit frustrating to say one thing and have someone else take it out of context and then argue that you said something entirely different.

> Atomic CAS is about 10x slower than a nonatomic CAS due to it requiring taking over cache and possibly invalidating it, and in a contested environment a CAS spinloop is even worse (not saying we _shouldn't_ do this, I was just half-sarcastically drawing comparison to a recent convo lol)

Ok quite curious about this since I've been told that CAS has similar performance impact to regular memory access and I thought this was the whole appeal of lockless concurrency. Can you provide a source or more reading for this? Not finding anything from a quick search

LordIdra:
Thanks and no worries

LordIdra:
Ok, interesting I didn't know that. Agree 100% that it's fine to do though - the pausing situation was mostly just me being confused that coroutines don't have an inbuilt way to do this, and I wanted to wait until the custom scheduler thing to be on the safe side that I'm writing performant code. I think the conversation ended up blowing up a bit in proportion to the issue lol. It seems necessary here anyway

Seggan:
O wait just read convo lol

LordIdra (changes requested):

PR #522, opened 2025-12-13
Procedural crafting additions

Intybyte:


Intybyte:
> Still confused on why this is needed over `entity.remove()`

what entity.remove, there is no remove method here, and it is a uuid not an entity class, contextualize a bit please

Seggan (approved):
nvm thought this was on EntityStorage

JustAHuman-xD (approved):

balugaq (approved):

PR #533, opened 2025-12-23
Add support for multiple invs

Seggan:
See the [Base PR](https://github.com/pylonmc/pylon-base/pull/501) for usage. 

Uhh yeah basically so the reason I couldn't use `addInventory` like with entities/fluid points is that `Inventory` is an interface that has multiple subclasses, so I can't deserialize it because they might be of different types.

### DIFF
--- a/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/base/PylonGuiBlock.kt
+++ b/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/base/PylonGuiBlock.kt
@@ -24,7 +24,7 @@ import xyz.xenondevs.invui.inventory.Inventory
 import xyz.xenondevs.invui.inventory.VirtualInventory
 import xyz.xenondevs.invui.inventory.event.UpdateReason
 import xyz.xenondevs.invui.window.Window
-import java.util.IdentityHashMap
+import java.util.*
 
 /**
  * A block that has an associated InvUI GUI that can be opened by right-clicking the block.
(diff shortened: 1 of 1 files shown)


pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/base/PylonGuiBlock.kt
```diff
@@ -90,46 +104,48 @@ interface PylonGuiBlock : PylonBreakHandler, PylonInteractBlock, PylonNoVanillaC
...
         val items = mutableListOf<ItemStack>()
         val invs = inventories[this] ?: return listOf()
-        for (inv in invs) {
+        for (inv in invs.values) {
             for (item in inv.items) {
                 item?.let(items::add)
             }
         }
         return items
     }
```
Intybyte: Some inventories might not need to be dropped

Intybyte (changes requested):

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/base/PylonGuiBlock.kt
Seggan: You can already just override onBreak and just not call super

LordIdra (approved):

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/block/base/PylonGuiBlock.kt
```diff
@@ -79,7 +93,7 @@ interface PylonGuiBlock : PylonBreakHandler, PylonInteractBlock, PylonNoVanillaC
     override fun onBreak(drops: MutableList<ItemStack>, context: BlockBreakContext) {
         guiBlocks.remove(this)
         val invs = inventories.remove(this) ?: return
-        for (inv in invs) {
+        for (inv in invs.values) {
             for (item in inv.unsafeItems) {
                 item?.let(drops::add)
             }
```
JustAHuman-xD: If we were concerned about dropping items we don't mean to, we could use an actual item update with a reason for block break.

I don't think its needed but I know it was discussed previously

JustAHuman-xD: Instead of accessing raw items get the item and use a setItem with a custom update reason, like done with cargo.

If it succeeds drop the item if it doesn't don't

LordIdra: Tbh it seems fine as is, it's a rare enough use case that it seems fine to just override and not drop

PR #542, opened 2025-12-30
Crucible additions changes whatever

Intybyte:


LordIdra (approved):

Seggan (changes requested):
This is exactly what `ItemTypeWrapper` is for. `ItemTypeWrapper.of(key).createItemStack()` go brr

LordIdra:
I merged the corresponding base PR and forgot it required this PR. So have just substituted `ItemTypeWrapper.of(key).createItemStack()` in master

Seggan:
It's redundant tho

Seggan:
The wrapper is designed for this sort of thing, namely getting items from namespacedkey. yours just has some extra post-processing added

Seggan:
I think you misunderstood what I meant. We don't need these utils in Core *at all*. They have some very specific fallbacks (which are undocumented btw) that are pretty much useful only in the cruicible

PR #543, opened 2025-12-31
Cargo polishing

LordIdra:
- Renames LogisticSlotType to LogisticGroupType
- Update cargo ticking logic to work with multiple target slots instead of just duplicating the items across all of them lmao
- Fix cargo routes not invalidating in certain situations
- Fix cargo routes with no endpoint not being cached (lol)
- Add logistic slots for vanilla inventories, and a method which gets all the logistic groups for a vanilla block

### DIFF
--- a/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/util/PylonUtils.kt
+++ b/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/util/PylonUtils.kt
@@ -323,39 +323,6 @@ fun findPylonItemInInventory(inventory: Inventory, targetItem: PylonItem): Int? 
     return null
 }
 
-/**
- * Compares two Pylon items to check if they have the same Pylon ID. If
- * neither item is a Pylon item, the material will be compared instead.
- *
- * @return Whether the items are the same.
- */
-fun ItemStack?.isPylonSimilar(item2: ItemStack?): Boolean {
-    // Both items null
-    if (this == null && item2 == null) {
-        return true
-    }
-
-    // One item null, one not null
-    if (!(this != null && item2 != null)) {
-        return false
-    }
-
-    val pylonItem1 = PylonItem.fromStack(this)
-    val pylonItem2 = PylonItem.fromStack(item2)
-
-    // Both pylon items null
-    if (pylonItem1 == null && pylonItem2 == null) {
-        return this.isSimilar(item2)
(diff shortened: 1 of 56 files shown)


pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/util/PylonUtils.kt
```diff
@@ -617,3 +640,41 @@ class MachineUpdateReason : UpdateReason
...
     = round(100 * block.type.getHardness() / block.getDestroySpeed(tool, true))
+
+fun getVanillaLogisticSlots(block: Block?): Map<String, LogisticGroup> {
+    val blockData = block?.state
+    return when (blockData) {
+        is Furnace -> mapOf<String, LogisticGroup>(
+            "input" to LogisticGroup(LogisticGroupType.INPUT, VanillaInventoryLogisticSlot(blockData.inventory, 0)),
+            "fuel" to LogisticGroup(LogisticGroupType.INPUT, FurnaceFuelLogisticSlot(blockData.inventory, 1)),
+            "output" to LogisticGroup(LogisticGroupType.OUTPUT, VanillaInventoryLogisticSlot(blockData.inventory, 2)),
+        )
```
OhmV-IR: these arbitrary names -> value mappings kinda scare me, if we use this a lot and ever change these names, it would be a real pain to change everywhere they're used (even worse if this is used by other plugins). An enum could be used to at least solve the arbitrary string keys problem, but why can't we just return a Set of logistic groups and avoid the problem alltogether?

LordIdra: Oh they're also used to display to the user which inventory is active when cycling between them. But that requires 'arbitrary' strings anyway

Seggan: > Oh they're also used to display to the user which inventory is active when cycling between them

We need them to be translatable then

pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/util/PylonUtils.kt
Seggan: Quite a few unused imports

Seggan (approved):
LGTM mostly

PR #552, opened 2026-01-04
Technical guide changes

LordIdra:
This PR prepares for the 'item menu' system I am planning to introduce. The central change is that buttons and pages are now more clearly separated. Previously, a page would have to supply an icon which could then be used by buttons linking to that page. This makes no sense for pages that do not have a button linking to them, which we are about to have a lot of, so I have changed pages to no longer have to supply an item for buttons linking to them.

Other changes:
- Slightly reorganises guide pages to conform better to the new format (and also remove the whole thing in `PylonCore` with settings pages being added in the `onEnable` method)
- Add shorthands to all pages to PylonGuide for ease of use

### DIFF
--- a/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/content/guide/PylonGuide.kt
+++ b/pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/content/guide/PylonGuide.kt
@@ -137,7 +137,7 @@ class PylonGuide(stack: ItemStack) : PylonItem(stack), PylonInteractor {
          */
         @EventHandler(priority = EventPriority.LOWEST)
         private fun join(event: PlayerJoinEvent) {
-            if (PylonConfig.pylonGuideOnFirstJoin && !event.player.hasPlayedBefore()) {
+            if (PylonConfig.PYLON_GUIDE_ON_FIRST_JOIN && !event.player.hasPlayedBefore()) {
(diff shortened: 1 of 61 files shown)


pylon-core/src/main/kotlin/io/github/pylonmc/pylon/core/content/guide/PylonGuide.kt
```diff
@@ -75,22 +84,52 @@ class PylonGuide(stack: ItemStack) : PylonItem(stack), PylonInteractor {
         val hiddenResearches: MutableSet<NamespacedKey> = mutableSetOf()
 
         @JvmStatic
-        val fluidsPage = FluidsPage()
+        val fluidsPage = object : SimpleDynamicGuidePage(
```
Seggan: `SimpleDynamicGuidePage` isn't abstract, you dont need an anonymous object

LordIdra:
> You've got a lost of assuming where keys are, ex assuming page names are under `pylon.<youraddon>.guide.page.<key>`. I would want a way to allow me to use a different component, for example dynamically generated pages

Could you elaborate on what changes you want here? I'm confused on what exactly you're referring to. I know PageButton does this, the idea was you'd override that or use your own page button if you were doing something fancy enough to want to supply your own components

Seggan (commented):
Oop just read the code, nvm

PR #594, opened 2026-02-13
Multi Event Handlers :D

JustAHuman-xD:
This is a huge PR, have fun guys, <3

- Adds MultiListeners & Universal/MultiHandlers for listening to multiple event priorities in the same method
- Also enables our item/block/entity interfaces to annotate themselves as well instead of being forced into the Normal priority with not ignoring cancelled
- Moves all rebar interface event handlers into companion objects of the interface to follow the newer standard instead of putting it all in a "RebarItemsListener" or "RebarBlockListener" etc
- Fixes or improves a few random things in the api
  - Redoes the RebarCopperBlock to use proper events & better prediction/support
  - BlockStorage#breakBlock now returns the items dropped rather than the item stacks
- Removes various bits of redundant code
  - for ex the manual subtracting of rebar blocks when placing or manual damaging of tools when breaking, they didnt actually do anything anymore since we no longer cancel the event so vanilla will do those things and the item in the players inventory gets overwritten by the event so the change we attempted didnt actually apply anyway.

I've tested this all in a basic sense and seems to be working, if a claims plugin has hardcoded prediction instead of blanket protection some blocks/items may still be permitted when they logically shouldnt be. We may want to look into *also* hardcoding some minimal support (which this also makes easier) but for now this vastly improves our api and support as it is.

### DIFF
--- a/rebar/src/main/kotlin/io/github/pylonmc/rebar/config/ConfigSection.kt
+++ b/rebar/src/main/kotlin/io/github/pylonmc/rebar/config/ConfigSection.kt
@@ -28,8 +28,8 @@ open class ConfigSection(val internalSection: ConfigurationSection) {
      * Gets all the values in the section that are themselves sections.
      * @throws NullPointerException if any top level keys do not correspond to a section
      */
-    fun getSections(): Set<ConfigSection> {
-        val configSections: MutableSet<ConfigSection> = mutableSetOf()
+    fun getSections(): List<ConfigSection> {
+        val configSections: MutableList<ConfigSection> = mutableListOf()
         for (key in internalSection.getKeys(false)) {
             configSections.add(getSection(key)!!)
         }
--- a/rebar/src/main/kotlin/io/github/pylonmc/rebar/config/adapter/ConfigAdapter.kt
+++ b/rebar/src/main/kotlin/io/github/pylonmc/rebar/config/adapter/ConfigAdapter.kt
@@ -7,6 +7,12 @@ import net.kyori.adventure.sound.Sound
 import org.bukkit.Bukkit
 import org.bukkit.NamespacedKey
 import org.bukkit.Registry
+import org.joml.Vector2d
+import org.joml.Vector2f
+import org.joml.Vector2i
+import org.joml.Vector3d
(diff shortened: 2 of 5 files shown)


rebar/src/main/kotlin/io/github/pylonmc/rebar/event/api/MultiListener.kt
```diff
@@ -0,0 +1,71 @@
...
+                ignoreCancelled = annotation.ignoreCancelled
+            } else if (method.isAnnotationPresent(UniversalHandler::class.java)) {
+                priorities = EventPriority.entries.toTypedArray()
+            } else {
+                continue
+            }
+
+            if (method.parameterTypes.size != 2 || !Event::class.java.isAssignableFrom(method.parameterTypes[0]) || method.parameterTypes[1] != EventPriority::class.java) {
+                throw IllegalStateException("Method ${method.name} in class ${this::class.java.name} must have exactly two parameters, the first being a subclass of Event, and the second being EventPriority")
+            } else if (!method.trySetAccessible()) {
```
OhmV-IR: just make this a separate if statement, else if is not needed since the throw will already exit out of the function. Helps to deobfuscate the flow a bit
```suggestion
            }
            if (!method.trySetAccessible()) {
```

OhmV-IR (commented):
lgtm except for these 2

rebar/src/main/kotlin/io/github/pylonmc/rebar/event/api/annotation/MultiHandler.kt
```diff
@@ -0,0 +1,101 @@
...
+
+                if (clazz.superclass != null) {
+                    findMethod(clazz.superclass, info)
+                } else {
+                    throw NoSuchMethodException("Could not find method $info.handlerMethod in class ${clazz.name} or its superclasses")
+                }
+            }
+        }
+
+        internal data class EventInfo(
```
OhmV-IR: @JvmSynthetic

OhmV-IR (commented):
oh some things

OhmV-IR (approved):

PR #606, opened 2026-02-23
Fix armor not stacking

OhmV-IR:
Closes https://github.com/pylonmc/pylon/issues/640

I know the fix is a bit... cursed but I don't see a way to be able to use the key of the item without refactoring the class heavily which I don't want to do.


JustAHuman-xD:
this is not how this should be fixed, and also introduces a new bug

OhmV-IR:
the attributes were overwriting each other because they had the same key, this fixes that by making the key depend on the armor slot.

JustAHuman-xD:
And items shouldn’t be able to overwrite attribute modifiers on other items that feels like it’s own bug

OhmV-IR:
uh no the item attributes don't overwrite each other directly, but minecraft tries to apply them to the player when they put the armor on, so that's when the overwriting happens.

JustAHuman-xD:
> also the tooltips are fine. image

Yeah thats not how the tooltip is supposed to look

PR #608, opened 2026-02-23
allow enchanting pylon items via anvil

### LINKED ISSUE #625: Lumber axe is not enchantable
AriaNight

OP
 — 1:21 PM
I cant put a mending book on my pylon lumber ax....

I put the book and ax in the anvil and for half a sec it shows then goes away and shows that I cant add it on.

OhmV-IR:
Closes https://github.com/pylonmc/pylon/issues/625


rebar/src/main/kotlin/io/github/pylonmc/rebar/recipe/RecipeListener.kt
```diff
@@ -231,14 +232,15 @@ internal object RebarRecipeListener : Listener {
...
-        if ((firstRebarItem != null && (secondItem != null && !secondItem.isEmpty)) || secondRebarItem != null) {
+        // Check if the result is a repaired item and if so, cancel it.
+        if(firstItem != null && e.result != null
+            && firstItem.itemMeta is Damageable && e.result!!.itemMeta is Damageable
+            && (firstItem.itemMeta as Damageable).damage < (e.result!!.itemMeta as Damageable).damage){
+            e.result = null
+            return
+        }
+        // Check if either input is a rebar item without VanillaAnvilItem and if the output is vanilla, cancel it
+        if((firstItem.isRebarAndIsNot<VanillaAnvilItem>() || secondItem.isRebarAndIsNot<VanillaAnvilItem>()) && RebarItem.fromStack(e.result) == null) {
```
Seggan: Formatting

Seggan (changes requested):

Seggan:
:point_up_2: was gonna mention that

Seggan:
...why are they clay balls again?

Intybyte:
> Hmm stone tools with gold model data would've been better. Also as ive mentioned like 598998498 times we can just cancel the piglin event

And as I have said 598998499 times there is the client showing the piglin picking the item up for half a second and dropping it over and over (client side prediction it was called iirc), but yeah I agree on the solution

Seggan:
I do not recall that...

Anyway yeah I think making them stone tools will solve the enchanting issue for bronze at least, among other things

LordIdra (approved):

PR #611, opened 2026-02-24
fix(guide): hide gameplay tooltips on unresearched guide items

SamuelJoseph23:
In ItemButton.kt, unresearched items had their model overridden to a Barrier block and a "not-researched" lore line prepended. However, the ItemStackBuilder retained the original item's metadata and tooltips, allowing players to view gameplay mechanic hints on locked items.

Fixed by chaining .clearLore() right after assigning the Barrier model, ensuring only the "Not enough research points" warning remains visible. Fixes pylonmc/pylon#660

**Verified by updating the rebar plugin and pasting into the pylon plugin folder and restarting server.**

Before Fix:

After Fix:


LordIdra (changes requested):
The issue is referencing  the hints section. So if you look at an unresearched item, currently there is a 'hints' section with instructions under it. We want to remove all the hints except for the 'shift left click to research' (and also the underlined 'hints' heading probably)

PR #612, opened 2026-02-24
Fix blockstorage dumpster fire

LordIdra:
- Fixes exceptions falling through to onDisable which prevented BlockStorage/EntityStorage from saving
- Fixes Rebar plugin disable being handled by addon disable listener (rebar cleanup should be done in onDisable)
- Isolates block serialization errors during cleanup
- Isolates entity serialization errors during cleanup

### DIFF
--- a/rebar/src/main/kotlin/io/github/pylonmc/rebar/Rebar.kt
+++ b/rebar/src/main/kotlin/io/github/pylonmc/rebar/Rebar.kt
@@ -403,7 +403,8 @@ object Rebar : JavaPlugin(), RebarAddon {
         logger.info("Loaded researches in ${(end - start) / 1000.0}s")
     }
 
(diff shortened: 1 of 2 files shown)


rebar/src/main/kotlin/io/github/pylonmc/rebar/Rebar.kt
```diff
@@ -403,11 +403,16 @@ object Rebar : JavaPlugin(), RebarAddon {
         logger.info("Loaded researches in ${(end - start) / 1000.0}s")
     }
 
-    override fun onDisable() {
-        PacketEvents.getAPI().terminate()
+    fun preDisable() {
```
Seggan:
```suggestion
    @JvmSynthetic
    internal fun preDisable() {
```

PR #617, opened 2026-02-25
balugaq/fix issues

balugaq:


Seggan (changes requested):
Fix works, but the sand should drop instead of break. All good otherwise.

balugaq:
QQ20260228-172503

Bronze anvil can't work on master branch actually...

balugaq:
> 2026-02-27.18-42-09.mp4

Fixed and tested

PR #618, opened 2026-02-26
Fix unresearched item dupe

balugaq:
Haven't tested yet


LordIdra (approved):
Ah OK I think I see. That's weird how that was even there in the first place. Cannot reproduce the issue locally to test that this branch removes it so will go off what you've said. Thanks for patching this.

PR #619, opened 2026-02-26
Add RebarConfigurableRecipesLoadedEvent

balugaq:


LordIdra (commented):
What is the reason for this event?

JustAHuman-xD:
> What is the reason for this event?

for systems that rely on recipes being loaded, I had a use case in occult operations

PR #621, opened 2026-02-27
Pipe and cargo facades

LordIdra:
image
image
image


Seggan (changes requested):
Bug

https://github.com/user-attachments/assets/ea554aee-c0ea-4524-bdca-7b6162f3d656

PR #625, opened 2026-02-28
Prevent break unresearched blocks

balugaq:


rebar/src/main/kotlin/io/github/pylonmc/rebar/block/BlockStorage.kt
```diff
@@ -223,6 +223,15 @@ object BlockStorage : Listener {
     fun isRebarBlock(block: Block): Boolean =
         (block.position.chunk.isLoaded) && get(block) != null
 
+    /**
+     * Returns whether the block at [location] is a Rebar block, or null if the
+     * chunk at [location] is not loaded
+     */
```
LordIdra: incorrect javadoc, never returns null

PR #631, opened 2026-03-01
add text color config adapter

OhmV-IR:
Adds an adapter that converts from a rgb config section, hsv config section or hex code to a TextColor. I was kinda shocked this wasn't already made xd.


OhmV-IR:
This is tested by https://github.com/pylonmc/pylon/pull/763

PR #643, opened 2026-03-04
Easy optimisations

LordIdra:
Just some small changes which will speed things up hopefully a lot
- Changed getSuppliedFluids to return a list of pairs, not a map - a map will probably be a lot slower since we are usually only dealing with like 1-3 elements
- Changed fluid ticking code to avoid repeated BlockStorage calls for the same block and avoid filtering fluid points by type every tick - this was taking up a lot of time in spark profiles
- Fix chunk unloading causing lots of errors due to BeaconDeactivateEvent being called


Seggan (approved):
LGTM

PR #659, opened 2026-03-08
Rewrite line wrapping + add stacks unit

LordIdra:
Summary of line wrapping changes: The old system folded the entire component into a string, did line wrapping, then turned it back into components. The new system does this in-place by building up lines incrementally by traversing the tree and appending components.


LordIdra:
@Seggan your eyes would be great here

PR #665, opened 2026-03-11
fix: jitpack build issues

### LINKED ISSUE #664: Jitpack build failed
https://jitpack.io/com/github/pylonmc/rebar/0.35.0/build.log

ybw0014:
resolves #664


ybw0014:
Set `archiveClassifier` to `null` would make the output of `jar` and `shadowJar` have the same name, and Gradle won't know which one to use.

The solutions are either:
- Remove `archiveClassifier` in shadowJar. This may cause other GH actions to fail to find the build artifact.
- Specify the task executed by jitpack. Jitpack uses `./gradlew clean assemble publishToMavenLocal` by default, which would execute both `jar` and `shadowJar`. Changing to `./gradlew clean jar publishToMavenLocal` will not execute `shadowJar`.

ybw0014:
done, will submit a PR to pylon as well

PR #688, opened 2026-03-23
Fix interfaces serialization/deserialization

Intybyte:
Fixes #687

### DIFF
--- a/rebar/src/main/kotlin/io/github/pylonmc/rebar/block/base/RebarCargoBlock.kt
+++ b/rebar/src/main/kotlin/io/github/pylonmc/rebar/block/base/RebarCargoBlock.kt
@@ -289,7 +289,6 @@ interface RebarCargoBlock : RebarLogisticBlock, RebarEntityHolderBlock {
             val block = event.rebarBlock
             if (block is RebarCargoBlock) {
                 cargoBlocks[block] = event.pdc.get(cargoBlockKey, RebarSerializers.CARGO_BLOCK_DATA)
-                    ?: error("Ticking block data not found for ${block.key}")
             }
         }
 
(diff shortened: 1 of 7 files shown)


Seggan (commented):
IMO we should warn if they are not found

rebar/src/main/kotlin/io/github/pylonmc/rebar/block/base/RebarCargoBlock.kt
```diff
@@ -289,15 +289,14 @@ interface RebarCargoBlock : RebarLogisticBlock, RebarEntityHolderBlock {
             val block = event.rebarBlock
             if (block is RebarCargoBlock) {
                 cargoBlocks[block] = event.pdc.get(cargoBlockKey, RebarSerializers.CARGO_BLOCK_DATA)
-                    ?: error("Ticking block data not found for ${block.key}")
```
LordIdra: I think this will insert null into `cargoBlocks` if `event.pdc.get(cargoBlockKey, RebarSerializers.CARGO_BLOCK_DATA)` returns null, so you should only operate on `cargoBlocks[block]` if the return value is not null. The reasoning being that with this behaviour, `x in cargoBlocks` will always return true after a block is loaded, which could cause some subtle bugs down the line because it's a very easy assumption to make and get wrong. (Same thing for others)

LordIdra (changes requested):

LordIdra:
> IMO we should warn if they are not found

The data in most/all of these interfaces is lazily initialised so it is quite common that the related data of a block may never have been accessed until it is unloaded and then reloaded again. That's intended behaviour, so a warning does not make sense

Seggan (approved):
Besides what Idra said LGTM

LordIdra:
I have fixed the issue I mentioned myself and also fixed an interface (and a half) that was missed

Seggan (approved):

PR #691, opened 2026-04-07
Docs improvements

LordIdra:
I just realised I forgot to PR this from like a month ago..


rebar/src/main/kotlin/io/github/pylonmc/rebar/util/ConfettiParticle.kt
```diff
@@ -14,6 +14,7 @@ import org.joml.Quaternionf
 import org.joml.Vector3f
 import java.util.Random
 import java.util.function.Consumer
+import kotlin.collections.filter
```
Seggan: Unused import?

Seggan (approved):

rebar/src/main/kotlin/io/github/pylonmc/rebar/util/ConfettiParticle.kt
LordIdra: no

LordIdra: wait yes. wtf, intellij does not display a warning... weird... anyway it's whatevers

PR #699, opened 2026-04-13
Make the recipes map linked

Seggan:
We don't ever register recipes async, and I would very much like the recipes to be stored in the order I registered them for the forming table


JustAHuman-xD:
~~and if an addon registers them async?~~

Seggan:
Dumb idea?

We can always synchronize

LordIdra (approved):

Intybyte (approved):

PR #706, opened 2026-04-17
Ghost block + multiblock component refactor

LordIdra:
Ghost blocks and multiblock components have become quite a mess with all the changes that have been made. This PR cleans it up by unifying the multiblock components under a single class (there was never really any good reason for it to be an interface) and adding an abstraction layer (`RebarGhostBlockHolder`) which separates the ghost block logic from the multiblock logic. The abstraction layer will also allow implementors of `RebarMultiblock` to spawn their own ghost blocks easily, which was previously very annoying to do. This also unifies handling of mixed multiblock components (rebar/vanilla) which has removed a lot of unnecessary logic.

### DIFF
--- a/rebar/src/main/kotlin/io/github/pylonmc/rebar/block/base/RebarSimpleMultiblock.kt
+++ b/rebar/src/main/kotlin/io/github/pylonmc/rebar/block/base/RebarSimpleMultiblock.kt
@@ -1,22 +1,11 @@
 package io.github.pylonmc.rebar.block.base
 
-import io.github.pylonmc.rebar.Rebar
 import io.github.pylonmc.rebar.block.BlockStorage
-import io.github.pylonmc.rebar.block.MultiblockCache
-import io.github.pylonmc.rebar.block.RebarBlock
-import io.github.pylonmc.rebar.block.RebarBlockSchema
 import io.github.pylonmc.rebar.datatypes.RebarSerializers
-import io.github.pylonmc.rebar.entity.EntityStorage
-import io.github.pylonmc.rebar.entity.RebarEntity
-import io.github.pylonmc.rebar.entity.display.BlockDisplayBuilder
-import io.github.pylonmc.rebar.entity.display.ItemDisplayBuilder
-import io.github.pylonmc.rebar.entity.display.transform.TransformBuilder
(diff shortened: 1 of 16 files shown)


rebar/src/main/kotlin/io/github/pylonmc/rebar/block/base/RebarSimpleMultiblock.kt
```diff
@@ -60,342 +40,98 @@ import kotlin.time.Duration.Companion.seconds
...
-            val schema = schema()
-            val display = ItemDisplayBuilder()
-                .itemStack(ItemStackBuilder.of(schema.material).addCustomModelDataString(key.toString()))
-                .glow(Color.WHITE)
-                .transformation(TransformBuilder().scale(0.5))
-                .build(block.location.toCenterLocation())
-            EntityStorage.add(MultiblockGhostBlock(display, key.toString()))
-            return display.uniqueId
+            @JvmStatic
+            fun of(vararg keys: NamespacedKey) = MultiblockComponent(
```
Seggan: This should account for me passing vanilla keys as well

Seggan (approved):

Seggan (changes requested):
Woops wrong button

Intybyte:
While this approach is much simpler, it also prevents people from making their own blocks using multiblocks, let's say i am a plugin dev and i want part of it to be done using idk, a Nexo or ItemsAdder block, now i can't add such feature without forking rebar unlike before

LordIdra:
> While this approach is much simpler, it also prevents people from making their own blocks using multiblocks, let's say i am a plugin dev and i want part of it to be done using idk, a Nexo or ItemsAdder block, now i can't add such feature without forking rebar unlike before

Yes you can, just extend MultiblockComponent

Intybyte:
> > While this approach is much simpler, it also prevents people from making their own blocks using multiblocks, let's say i am a plugin dev and i want part of it to be done using idk, a Nexo or ItemsAdder block, now i can't add such feature without forking rebar unlike before
> 
> Yes you can, just extend MultiblockComponent

oh it is an open class, then it should be fine

Intybyte (approved):

PR #707, opened 2026-04-17
NUCLEAR OBLITERATION OF ENEMY

LordIdra:
FIXES MY ETERNAL NEMESIS https://github.com/pylonmc/pylon/issues/955

this should finally fix the issue where random 2-length cargo duct displays would stay floating around, and also fix the errors that people frequently observed in console regarding cargo


Seggan (approved):
Haven't tested but I'll trust you

PR #714, opened 2026-04-25
Fix breaking rebar blocks doesn't damage the tool

### LINKED ISSUE #713: Breaks rebar blocks doesn't damage the tool in hand

balugaq:


balugaq:
QQ_1777124335619

As the picture shows, when player breaks a block, it calls `ServerPlayerGameMode.destoryBlock` and `callEvent (L352)` and the event is handled by `BlockListener`, goes to `BlockStorage.removeBlock`, then set the block to air (`PlayerBreak` 
 doesn't override `shouldSetToAir` so returns `true` by default), causing the failure of the check at `ServerPlayerGameMode (L372)`.
Normally the block is not air, and goes to the else branch, then calls `mineBlock (L401)` to damage the tool.

So add a `override` to fix this problem.

**tested*

LordIdra (approved):

PR #715, opened 2026-04-28
Recipe matching system rewrite

SchnTgaiSpock:
This was originally supposed to be the companion to https://github.com/pylonmc/pylon/pull/962 to let damaged items be smelted, but then i decided to make it more general for any component, and now i wanna just overhaul the recipe matching in `RecipeListener`

As of writing, this branch works with https://github.com/pylonmc/pylon/pull/962

###### Also this is my first time writing kotlin so apologies if its absolute dogsh*t


rebar/src/main/kotlin/io/github/pylonmc/rebar/recipe/RecipeInput.kt
```diff
@@ -28,7 +30,14 @@ sealed interface RecipeInput {
...
 
-        operator fun contains(itemStack: ItemStack): Boolean = ItemTypeWrapper(itemStack) in items
+        operator fun contains(itemStack: ItemStack): Boolean {
+            for (item in items) {
+                if (item.createItemStack().matchesWithoutData(itemStack, ignoreComponents, true)) {
+                    return true
+                }
+            }
+            return false
+        }
```
JustAHuman-xD, line 42: Not sure how I feel about creating an itemstack just to get it garbage collected this many times potentially.

Maybe we could have a cache or something so that there's just one immutable instance used or something but yeah.

SchnTgaiSpock: made it so rebar items match against the template item but the vanilla item is a record so i cant put a cache field, do you think its worth it to derecord it

SchnTgaiSpock:
okayyy so, slight problem, since steel boots and palladium boots both use diamond boots, mc gets the recipe (diamond_boots -> palladium_nugget) so when mc thinks steel boots smelt into platinum nuggets and wont even bother calling fuel burn or start smelt events (which is where we would normally correct it) if theres something in the output 
image

SchnTgaiSpock:
you can never hate mojang enough bro

SchnTgaiSpock:
everythings ready for review, the thing above i think is a minor enough edge case ithink its fine to leave in cuz i genuinely dont know what to do with it besides ticking the furnace itself

rebar/src/main/kotlin/io/github/pylonmc/rebar/recipe/vanilla/Cooking.kt
```diff
@@ -51,47 +56,66 @@ sealed class CookingRecipeWrapper(final override val recipe: CookingRecipe<*>) :
...
+        ingredient.ignoreComponents.add(DataComponentTypes.DAMAGE)
+    }
     val result = config.getOrThrow("result", ConfigAdapter.ITEM_STACK)
-    val recipe = cons(key, result, ingredient.asRecipeChoice(), experience, cookingTime)
+    val recipe = recipeCons(key, result, ingredient.asRecipeChoice(), experience, cookingTime)
     config.get("category", ConfigAdapter.ENUM.from<CookingBookCategory>())?.let { recipe.category = it }
     config.get("group", ConfigAdapter.STRING)?.let { recipe.group = it }
-    return recipe
+    return wrapperCons(recipe, ingredient)
 }
```
LordIdra, line 109: Could we make this not generic (i.e. not take in the constructors as parameters)? This is extremely hard to read as-is, I think it would be a lot more readable split up even though it's more code

rebar/src/main/kotlin/io/github/pylonmc/rebar/recipe/vanilla/Cooking.kt
```diff
@@ -4,23 +4,28 @@ import io.github.pylonmc.rebar.config.ConfigSection
...
 import org.bukkit.Material
 import org.bukkit.NamespacedKey
 import org.bukkit.inventory.*
 import org.bukkit.inventory.recipe.CookingBookCategory
 import xyz.xenondevs.invui.gui.Gui
 
-
-sealed class CookingRecipeWrapper(final override val recipe: CookingRecipe<*>) : VanillaRecipeWrapper {
-    override val inputs: List<RecipeInput> = listOf(recipe.inputChoice.asRecipeInput())
+sealed class CookingRecipeWrapper(final override val recipe: CookingRecipe<*>, val recipeInput: RecipeInput.Item) : VanillaRecipeWrapper {
```
LordIdra, line 22: What's the purpose of the new recipeInput parameter? I can't find where it's used if anywhere?

rebar/src/main/kotlin/io/github/pylonmc/rebar/recipe/vanilla/Crafting.kt
```diff
@@ -56,13 +61,24 @@ class ShapedRecipeWrapper(override val recipe: ShapedRecipe) : CraftingRecipeWra
...
-sealed class AShapelessRecipeWrapper(recipe: CraftingRecipe) : CraftingRecipeWrapper(recipe) {
+    fun matches(items: List<ItemStack?>): Boolean {
+        if (RebarRecipe.matchesShaped(items, recipeInput, 3, 3)) return true
+        val mirror = items.toMutableList()
+        while (mirror.size < 9) mirror.add(null)
+        mirror[2] = mirror[0].also{ mirror[0] = mirror[2] }
+        mirror[5] = mirror[3].also{ mirror[3] = mirror[5] }
+        mirror[8] = mirror[6].also{ mirror[6] = mirror[8] }
+        return RebarRecipe.matchesShaped(mirror, recipeInput, 3, 3)
+    }
```
LordIdra, line 67: I believe that shaped recipes are automatically mirrored already when added to the server. Or is this because we are manually reimplementing matchesShaped? If so, this should really be in matchesShaped, not in here

rebar/src/main/kotlin/io/github/pylonmc/rebar/util/LRUCache.kt
```diff
@@ -0,0 +1,9 @@
+package io.github.pylonmc.rebar.util
+
+class LRUCache<K, V>(private val initialCapacity: Int) :
+    LinkedHashMap<K, V>(initialCapacity, 0.75f, true) {
+
+    override fun removeEldestEntry(eldest: MutableMap.MutableEntry<K, V>?): Boolean {
+        return size > initialCapacity
+    }
+}
```
LordIdra: Needs documenting and maybe should also be internal

rebar/src/main/kotlin/io/github/pylonmc/rebar/recipe/RebarRecipe.kt
LordIdra, line 1: A lot of the code in this file is incredibly cursed, e.g. the manual re-implementation of shaped crafting, searchRecipes (which I think is only used by the vanilla wrappers?), etc. I would much prefer to keep this file as clean as possible because it's an important public interface. Can we move this code closer to where it is being used to 1) clean up this file and 2) make it clearer why half of these methods exist? It took me a solid 10-15 minutes just to understand why we were manually re-implementing shaped crafting partially because of the indirection, having this logic in e.g. RecipeListener would be good

rebar/src/main/kotlin/io/github/pylonmc/rebar/recipe/RebarRecipe.kt
```diff
@@ -46,12 +50,118 @@ interface RebarRecipe : Keyed {
...
+        private fun cacheKey(hash: Int, recipeType: RecipeType<*>) = 31 * hash + recipeType.hashCode()
+
+        fun getCached(hash: Int, recipeType: RecipeType<*>) = cache[cacheKey(hash, recipeType)]
+        fun isCached(hash: Int, recipeType: RecipeType<*>) = cacheKey(hash, recipeType) in cache
+        fun cache(hash: Int, recipeType: RecipeType<*>, recipe: RebarRecipe?) {
+            cache[cacheKey(hash, recipeType)] = recipe
+        }
+        fun clearCache() { cache.clear() }
+
+        inline fun <reified T : RebarRecipe> searchRecipes(recipeType: RecipeType<out T>, hint: NamespacedKey?, hash: Int, pred: Predicate<T>): T? {
```
LordIdra: Needs documenting

rebar/src/main/kotlin/io/github/pylonmc/rebar/recipe/RebarRecipe.kt
```diff
@@ -46,12 +50,118 @@ interface RebarRecipe : Keyed {
...
             }
+
+        private fun cacheKey(hash: Int, recipeType: RecipeType<*>) = 31 * hash + recipeType.hashCode()
+
+        fun getCached(hash: Int, recipeType: RecipeType<*>) = cache[cacheKey(hash, recipeType)]
+        fun isCached(hash: Int, recipeType: RecipeType<*>) = cacheKey(hash, recipeType) in cache
+        fun cache(hash: Int, recipeType: RecipeType<*>, recipe: RebarRecipe?) {
+            cache[cacheKey(hash, recipeType)] = recipe
+        }
+        fun clearCache() { cache.clear() }
```
LordIdra: How would caching work in practice for Pylon recipes? Could you maybe demonstrate it by adding a cache to a few Pylon recipes so we can see what this looks like from a user perspective?

LordIdra (changes requested):
-

LordIdra:
Another thing I should flag, recipe autocompletion from the crafting book is very likely to be broken by these changes and will need testing

rebar/src/main/kotlin/io/github/pylonmc/rebar/recipe/vanilla/Cooking.kt
SchnTgaiSpock: its used in the `matches` method which needs the `RecipeInput.Item` instead of `RecipeChoice` to properly match rebar items w/out durability (or any other component) for reasons outlined here https://github.com/pylonmc/pylon/pull/962#issuecomment-4284479069

rebar/src/main/kotlin/io/github/pylonmc/rebar/util/LRUCache.kt
SchnTgaiSpock: Added docs; why internal tho? i think its a useful data structure

rebar/src/main/kotlin/io/github/pylonmc/rebar/recipe/RebarRecipe.kt
SchnTgaiSpock: They're pretty generic util methods so I put them in a separate util class, also added docs to the listener

rebar/src/main/kotlin/io/github/pylonmc/rebar/recipe/RebarRecipe.kt
SchnTgaiSpock: Created companion pr https://github.com/pylonmc/pylon/pull/1011

LordIdra:
Sorry for the lack of activity on this

This PR is probably one of the most dense/complex ones we've had on here and has a lot of  weird behaviour, so much so that I still struggle to understand everything that's going on despite having written most of the original code and having looked over this several times. It's not something I would want to merge without the benefits being very worth it. I'll dig into the problems at some point and see if I can figure out any simpler solutions. Will probably have another look when I try to abstract out recipe processing as that may help here

SchnTgaiSpock:
The first issue I ran into for [pylon#849](https://github.com/pylonmc/pylon/issues/849) was that minecraft only gives you two choices for specifying a recipe - exact match or material match, and with exact match then items that took dura damage wouldn't be able to be used in recipes.

This means we have to use material match, but now we have to block other items using the same material, e.g. putting a diamond sword (material for steel sword) in a furnace

In the onFuelBurn event for example, it doesn't provide us the recipe that triggered it so we have to do a search in our registered recipes to make sure the diamond sword can actually be put in a furnace. It can't, so the event is cancelled.

For other crafting and crafting related events, if the event object gives us a recipe key, we have to get the `RebarRecipe` from the key and check if it can be crafted with the provided ingredients. If not then we need to do another search.

There's a lot of searching done so that's what the lru cache in `RecipeService` is for, its indexed by a combination of the ingredients' hash and recipe type. `searchRecipes` is just a util that combines checking the recipe key from mc events, checking the cache, searching the `RecipeType` provided, and caching the result. The shaped/shapeless matching methods are for these searches.

I needed to add the recipe inputs to the vanilla recipe wrappers because its not possible to get the RecipeInput.Item back from a MaterialChoice if it's a rebar item.

LordIdra:
Great, thanks for the detailed explanation - I will definitely look again at some point

LordIdra:
Superseded by https://github.com/pylonmc/rebar/pull/835

Thanks for the PR - we did re-use some parts in #835.

PR #720, opened 2026-05-01
Add join handler

balugaq:
Companion to https://github.com/pylonmc/pylon/pull/970


Seggan (approved):

LordIdra (changes requested):
Small thing but why is this only called for non-first-time-players? If a player spawns with an item in their inventory e.g. as part of a kit surely we'd want to call this then as well, deliberately excluding this as an edge case seems redundant

LordIdra (approved):

PR #721, opened 2026-05-02
fix primitives checks

### LINKED ISSUE #701: Using ConfigAdaptor.INTEGER to get a double does not throw an error
example
```
bruh: 0.2
```
getOrThrow("bruh", ConfigAdapter.INTEGER);

this should throw an exception

balugaq:


Seggan (commented):
IMO we *should* allow coercion of floats to integers. @pylonmc/devs any thoughts?

OhmV-IR:
> IMO we *should* allow coercion of floats to integers. @pylonmc/devs any thoughts?

I don't think we should for anything since then we lose the self-documenting type feature and it's unclear to server admins what type should be in the confit field. 
Eg:
myIntField: 3.0

An admin might see this and think they can change the field to any float and it will work because it is already a float, but when they try it will floor it and not work, leading to a subtle bug.

rebar/src/main/kotlin/io/github/pylonmc/rebar/config/adapter/ConfigAdapter.kt
```diff
@@ -141,6 +141,24 @@ interface ConfigAdapter<T> {
...
+        private inline fun <reified T : Number> numberAdapter(
+            min: T,
+            max: T,
+            crossinline toType: Number.() -> T
+        ): ConfigAdapter<T> = ConfigAdapter { value ->
+            if (value is String) return@ConfigAdapter value.toDouble().toType()
+
+            check (value is Number) { "Expect ${Number::class.java}, got ${value::class.java}" }
+            if (value is Float || value is Double) {
+                check(min is Float || min is Double) { "Expect ${min::class.java}, got ${value::class.java}" }
```
Seggan:
```suggestion
                check(min is Float || min is Double) { "Expected ${T::class.java.canonicalName}, got ${value::class.java.canonicalName}" }
```
And likewise

rebar/src/main/kotlin/io/github/pylonmc/rebar/config/adapter/ConfigAdapter.kt
```diff
@@ -141,6 +141,24 @@ interface ConfigAdapter<T> {
...
+            }
+
+            if (min !is Double) {
+                check(min.toDouble() <= value.toDouble() && value.toDouble() <= max.toDouble()) { "Expect ${min::class.java}, got ${value::class.java}" }
```
Seggan:
```suggestion
                check(value.toDouble() in (min.toDouble())..(max.toDouble())) { "Expected value between $min and $max, got $value" }
```

Seggan (changes requested):

LordIdra (changes requested):
Is there a reason for the new numberAdapter abstraction? I would have thought that max/min values are a rare edge case which I'd hope would be caught by kotlin anyway and not overflow, and the logic can be much more simply just hardcoded into the relevant converters

balugaq:
> Is there a reason for the new numberAdapter abstraction? I would have thought that max/min values are a rare edge case which I'd hope would be caught by kotlin anyway and not overflow, and the logic can be much more simply just hardcoded into the relevant converters

QQ_1778398732011

Kotlin cannot handle this case. It is essential to check the max/min values edges, or use something like `toIntExact`.

LordIdra:
> Kotlin cannot handle this case. It is essential to check the max/min values edges, or use something like `toIntExact`.

Ah I see, that's annoying

LordIdra (approved):

PR #724, opened 2026-05-03
Allow to modify/extend search algorithm

### LINKED ISSUE #723: Allow to modify/extend search algorithm
Unlike English, Chinese characters have multiple indexing methods for retrieval. And it happens in not only Chinese but also the following languages:

> 1. Japanese
> Like Chinese, Japanese has Kanji, Hiragana, Katakana, Romaji (spaced/unspaced) and initial abbreviation. The same word has multiple valid writing forms, causing search mismatch if only exact original text is supported.
> 2. Korean
> Korean uses Hangul with multiple Romanization standards. One word can be written in spaced, unspaced Roman letters or initial abbreviations, which brings the same multi-index search problem.
> 3. Thai
> Thai has native script and Royal Thai Romanization. Variations in spacing, vowel omission and initial shorthand lead to many valid search forms for one single word.
> 4. Vietnamese
> Vietnamese has diacritical tones. Users often search without tones, or use shorthand / unspaced spelling. Different tone marks and casual writing styles cause search inconsistency.
> 5.  Arabic / Persian / Urdu
> They have native Arabic script and multiple Latin transliterations. People often omit diacritics or use connected writing, resulting in many search variants for one term.
> 6. Russian & Cyrillic languages
> Including Ukrainian, Belarusian.
> They use Cyrillic alphabet with various Latin transliteration schemes. Users commonly search with Latin letters or initials instead of original Cyrillic text.
> Generated by Doubou

Take Chinese for instance:

| String | Full Spell (With Space) | Full Spell (No Space) | First Letter Abbreviation |
| --- | --- | --- | --- |
| 石头 | shi tou | shitou | st |
| 宝石 | bao shi | baoshi | bs |
| ... | ... | ... | ... |

If a user searches for `蓝宝石`, relevant results should be matched by the keyword `宝石`, its full pinyin `baoshi`, or its abbreviation `bs`.
Currently, the system only supports matching by the original Chinese character `宝石`. Searching by Pinyin baoshi or initials bs cannot retrieve `蓝宝石` at all.
Therefore, allow other addons to customize and extend the built-in search algorithm not only greatly improves playing experience but also is necessary and essential.

balugaq:
Closes #723


Seggan (approved):

rebar/src/main/kotlin/io/github/pylonmc/rebar/content/guide/RebarGuide.kt
```diff
@@ -143,10 +145,24 @@ class RebarGuide(stack: ItemStack) : RebarItem(stack), RebarInteractor {
...
         @JvmStatic
-        val searchItemsAndFluidsPage = SearchItemsAndFluidsPage()
+        val latinSearchItemsAndFluidsPage = LatinSearchItemsAndFluidsPage()
 
         @JvmStatic
-        val searchItemsAndFluidsButton = PageButton(Material.OAK_SIGN, searchItemsAndFluidsPage)
+        val latinSearchItemsAndFluidsButton = PageButton(Material.OAK_SIGN, latinSearchItemsAndFluidsPage)
+
+        @JvmStatic
+        val localSearchPages: MutableMap<Locale, PageButton> = mutableMapOf()
```
LordIdra, line 167: A doc on here wouldn't hurt, to explain why there are multiple search pages

LordIdra (approved):

LordIdra (changes requested):
Just doing a second pass on this, how would you go about actually implementing different search logic for a locale? Surely we'd want to have that functionality as part of Rebar and not an addon which this seems to be tailored towards? Can the functionality not just be coded directly into the search page so that the search page screen behaves differently for some specific locales, instead of having per-locale pages?

balugaq:
> Just doing a second pass on this, how would you go about actually implementing different search logic for a locale? Surely we'd want to have that functionality as part of Rebar and not an addon which this seems to be tailored towards? Can the functionality not just be coded directly into the search page so that the search page screen behaves differently for some specific locales, instead of having per-locale pages?

I have done something about it. Take Chinese as an example, it's [Pinyin lib](https://github.com/houbb/opencc4j) has a size of 3.0MB, which occupies a considerable amount of space yet achieves a very minor function.

LordIdra:
> I have done something about it. Take Chinese as an example, it's [Pinyin lib](https://github.com/houbb/opencc4j) has a size of 3.0MB, which occupies a considerable amount of space yet achieves a very minor function.

I don't understand. How does that relate to the PR or the question?

balugaq:
> > I have done something about it. Take Chinese as an example, it's [Pinyin lib](https://github.com/houbb/opencc4j) has a size of 3.0MB, which occupies a considerable amount of space yet achieves a very minor function.
> 
> I don't understand. How does that relate to the PR or the question?

In my opinion, it's not worthwhile to import a huge prebuilt library into Rebar just to implement this single function.
So add APIs for external plugins to implement the feature. Server owners can choose to install the supporting library based on whether relevant players exist on their servers.

Seggan:
*As far as I can tell* this would allow external addons to return a different search page with different search rules based on locale.

LordIdra:
It does not seem like a good idea to require servers to install a separate addon to get proper search functionality in certain languages in a plugin that has per-client translation; that is not a good solution. Can we not in-house the logic? I don't understand quite why we need this library looking over the existing issue

LordIdra:
> It does not seem like a good idea to require servers to install a separate addon to get proper search functionality in certain languages in a plugin that has per-client translation; that is not a good solution. Can we not in-house the logic? I don't understand quite why we need this library looking over the existing issue

Hi @balugaq any update?

balugaq:
> > It does not seem like a good idea to require servers to install a separate addon to get proper search functionality in certain languages in a plugin that has per-client translation; that is not a good solution. Can we not in-house the logic? I don't understand quite why we need this library looking over the existing issue
> 
> Hi @balugaq any update?

Oh sorry, my email didn't notice me.

Just as I said some of the libs to support the search may too big (MB-level) and there may no relative players to active this part of search. 

But what you said also makes sense. The content of this PR also supports the way you mentioned. As for whether to add additional plugins to support related functions, can we open a discussion?

LordIdra:
I'm still really not sure how to deal with this issue. As you say adding a 3MB library for one language probably is not an acceptable solution, but any solution we implement should really ideally be universal. Is there any middle ground here that would catch 90%+ of cases? And how important is this to implement - I understand that the issue with chinese specifically is matching on synonyms, but how much is that an issue in practice? In english there are also synonyms but they can be ignored when searching and the search function still works fine, is it different in chinese?

PR #727, opened 2026-05-03
Configurable bossbar color and overlay

### LINKED ISSUE #673: Configurable Vanilla Waila Color
**Issue reported by: overlordidra:**
like gray

balugaq:


JustAHuman-xD:
This feels like something that shouldn't be up to the player tbh

LordIdra:
> This feels like something that shouldn't be up to the player tbh

Agreed

Justin elaborated in vc that this would be an issue for resource packs that use custom bossbars, so this should if anything should be a per-server config rather than per-client

Seggan:
Agree

PR #744, opened 2026-05-12
Small Bug Fixes & Features

### LINKED ISSUE #455: Pylon Item interface for when picked up

### LINKED ISSUE #503: Add whenEntityLoads overloads for non-Pylon entities
This would be very useful to have (I've run into a scenario where it would be very useful, and I think we will need it in the future for similar scenarios).

Potential implementation is effectively the same as the current implementation for Pylon entities, with a few tweaks. Potentially using the exact same fields but I'm not sure.

Feels a bit weird to put this in EntityStorage but not sure where else it should go, so that's probably fine.

### LINKED ISSUE #550: A strange search bug
https://github.com/user-attachments/assets/9f5832e6-71ca-42c3-8391-fe19c17e169a

Reproduce:
1. seacrh anything
2. close gui
3. clear text box and search the same thing again

then the word seems to be recognized as the empty value.

Expected:
it should normally search instead of the `empty box state`

### LINKED ISSUE #585: Guide page items have custom model data added repeatedly
Image

### LINKED ISSUE #622: add guide hint for shift+right click to go back to main page

### LINKED ISSUE #633: Pipe intersection point's lack custom model data
**Issue reported by: pandicka:**
probably can fix it with all cardinal points as model data but that's not optimal wise I'd guess

### LINKED ISSUE #708: Remove Rebar from addon list in bStats

### LINKED ISSUE #735: BlockStorage should store it's pdc type
instead of recreating it every save & load call (listTypeFrom)

### LINKED ISSUE #740: Show inner item in waila when looking at a Item Entity
Image

### LINKED ISSUE #747: Middle clicking a Rebar block does not put the item in the hand if the block already exists in the inventory

JustAHuman-xD:


Seggan (approved):

rebar/src/main/kotlin/io/github/pylonmc/rebar/item/RebarItemListener.kt
```diff
@@ -218,14 +209,6 @@ internal object RebarItemListener : Listener {
...
         event.sourceSlot = newSourceSlot
         event.targetSlot = event.player.inventory.heldItemSlot
-
-        // don't question this idk wtf is going on - seems we have to manually do the swap in the hotbar
-        if (sourceSlot <= 8) {
-            val source = event.player.inventory.getItem(event.sourceSlot)
-            val target = event.player.inventory.getItem(event.targetSlot)
-            event.player.inventory.setItem(event.sourceSlot, target)
-            event.player.inventory.setItem(event.targetSlot, source)
-        }
```
LordIdra, line 228: why was this removed? did you figure out what it did and that it's redundant? this comment was sponsored by @Intybyte

JustAHuman-xD: the comment described what it was supposed to do, manually move the items, but in testing it wasnt needed

JustAHuman-xD: even without it everything worked perfectly

Intybyte (approved):
fine beside what idra said, didn't test

rebar/src/main/kotlin/io/github/pylonmc/rebar/item/RebarItemListener.kt
LordIdra: are you absolutely sure you've tested all scenarios? I remember there were a crazy number of edge cases with this function which I kept discovering, I'd rather not delete anything here unless we're absolutely sure it isn't necessary

JustAHuman-xD: I tested as many as I could think of, either way its definitely *better* for us to remove this code because we were breaking the rules of the event & its assumptions by having it there.

We basically told the event, hey this is the source and this is the target, under the assumption the event would move it, but then *we* move it, making the source the target and the target the source, but we don't update the events data when we do that. So if the event does whats its supposed to afterwards, it will move it right back so no actual change would happen.

LordIdra: ok, makes sense

rebar/src/main/kotlin/io/github/pylonmc/rebar/item/base/RebarPickupable.kt
LordIdra, line 1: lmao

LordIdra (approved):

PR #736, opened 2026-05-10
Rework Untranslating Items & ArmorTextureEngine

JustAHuman-xD:
Should fix the creative mode inventory overriding data on the actual items with the translated version as well as a number of bugs with how the previous version actually worked

Also reworks the ArmorTextureEngine to be based on NMS (for performance) & to properly reset as well

PR #751, opened 2026-05-17
Replace Entity Lib & PacketEvents

JustAHuman-xD:
(relies on #736 as it removes the other packet events usage)

Removes the dependency on EntityLib & PacketEvents ( fixes #689 ) in favor of a custom packet entity implementation using NMS
Now that its custom: 
- its more flexible
- we can fix anything without having to rely on the external plugins
- the front end API is in normal bukkit classes so its MUCH more user friendly
- now that its an interface people can create their own implementations for even *more* flexibility

### DIFF
--- a/nms/src/main/kotlin/io/github/pylonmc/rebar/nms/entity/BlockTextureEntityImpl.kt
+++ b/nms/src/main/kotlin/io/github/pylonmc/rebar/nms/entity/BlockTextureEntityImpl.kt
@@ -1,0 +1,317 @@
+package io.github.pylonmc.rebar.nms.entity
+
+import com.mojang.math.MatrixUtil
+import io.github.pylonmc.rebar.Rebar
+import io.github.pylonmc.rebar.block.RebarBlock
+import io.github.pylonmc.rebar.culling.BlockCullingEngine
+import io.github.pylonmc.rebar.entity.display.transform.TransformUtil.toMatrix
+import io.github.pylonmc.rebar.entity.packet.BlockTextureEntity
+import io.github.pylonmc.rebar.entity.packet.BlockTextureEntity.Companion.MAX_SCALE_INCREASE
+import io.github.pylonmc.rebar.entity.packet.BlockTextureEntity.Companion.SCALE_DISTANCE_INCREASE
+import io.github.pylonmc.rebar.entity.packet.BlockTextureEntity.Companion.SCALE_DISTANCE_THRESHOLD
+import io.github.pylonmc.rebar.util.delayTicks
+import kotlinx.coroutines.launch
+import net.minecraft.network.protocol.Packet
+import net.minecraft.network.protocol.game.ClientboundAddEntityPacket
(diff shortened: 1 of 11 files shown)


rebar/src/main/kotlin/io/github/pylonmc/rebar/block/RebarBlock.kt
```diff
@@ -143,42 +139,30 @@ open class RebarBlock private constructor(val block: Block) : Keyed {
...
-        )
-        meta.width = 0f
-        meta.height = 0f
-    }
-
-    /**
-     * Use this method to make any changes to the block texture entity, such as changing its item,
-     * transformation, etc, after initialization. (see [setupBlockTexture])
-     */
-    protected fun updateBlockTexture(updater: (BlockTextureEntity, ItemDisplayMeta) -> Unit) {
```
Seggan, line 177: Why'd you nuke this?

rebar/src/main/kotlin/io/github/pylonmc/rebar/culling/BlockCullingEngine.kt
```diff
@@ -61,8 +62,9 @@ object BlockCullingEngine : Listener {
 
     internal val occludingCache = mutableMapOf<UUID, MutableMap<Long, ChunkData>>()
 
-    internal val blockTextureOctrees = mutableMapOf<UUID, Octree<RebarBlock>>()
-    internal  val culledBlockOctrees = mutableMapOf<UUID, Octree<RebarBlock>>()
+    @get:ApiStatus.Internal
+    val blockTextureOctrees = mutableMapOf<UUID, Octree<RebarBlock>>()
+    internal val culledBlockOctrees = mutableMapOf<UUID, Octree<RebarBlock>>()
```
Seggan, line 67: Synthetic

rebar/src/main/kotlin/io/github/pylonmc/rebar/block/RebarBlock.kt
JustAHuman-xD: The reason it was a thing was to provide easy access to the entity & meta, because how to use it isn't clear if you weren't familiar with EntityLib, but now that its our own, and everything is just within the entity itself it wasn't necessary

Intybyte (approved):

nms/src/main/kotlin/io/github/pylonmc/rebar/nms/entity/BlockTextureEntityImpl.kt
```diff
@@ -0,0 +1,328 @@
...
+     */
+    private var scaleTranslation = Vector3f()
+        get() {
+            if (System.currentTimeMillis() - translationTimestamp > 5000) {
+                field = calculateScaleTranslation()
+                translationTimestamp = System.currentTimeMillis()
+            }
+            return field
+        }
+    private var translationTimestamp = 0L
```
LordIdra, line 85: translation changes start point of all transformations not just scaling, so misleading names

LordIdra (approved):

PR #770, opened 2026-05-21
Furance Recipe Cache Fix

### LINKED ISSUE #602: Furnace only smelts one tin at a time

JustAHuman-xD:
fixes #602 & some other bugs that no one had caught yet lol


Intybyte (commented):
Good documentation, can we also get something in the package-info.java? 

Right now there is some docs about the crafting autocompletion fix, but now that in the same package there is also stuff for handling furnace logic for the sake of completeness something about this should be written there

Seggan (approved):

LordIdra (approved):

PR #774, opened 2026-05-24
fix: resolve RebarGuide companion object accessibility for Kotlin compileOnly dependencies

mc506lw:
## Summary
Fix `Unresolved reference` compilation errors when Kotlin projects access `RebarGuide` static API via `compileOnly` dependency.

## Problem
Kotlin-based Rebar addons (using `compileOnly` dependency) cannot call any `@JvmStatic` methods in `RebarGuide.Companion`:

```kotlin
// ❌ Compilation error: Unresolved reference 'getRootPage'
val page = RebarGuide.getRootPage()

// ❌ Unresolved reference 'hideItem'
RebarGuide.hideItem(someKey)
```

**Affected scope:** All `@JvmStatic` members (methods + properties)  
**Not affected:** Java projects, other Rebar APIs like `RebarItem.register()`

## Root Cause
The `Companion object : Listener` pattern creates complex bytecode that confuses Kotlin's metadata resolution:

```kotlin
// Current (problematic)
class RebarGuide(...) : RebarItem(...), RebarInteractor {
    companion object : Listener {  // ← Unusual: companion implements interface
        @JvmStatic val rootPage = RootPage()
        @EventHandler  // ← Event handler in companion
        private fun join(event: PlayerJoinEvent) { ... }
        // ... 25 more @JvmStatic members
    }
}
```

This combination triggers a Kotlin compiler edge case where `@JvmStatic` metadata becomes inaccessible from external modules.

## Solution
**Extract event handling into a dedicated listener object** (follows existing patterns like `BlockListener`):

### Changes
- **[RebarGuide.kt](path/to/RebarGuide.kt)**
  - Remove `: Listener` from companion object
  - Create internal `object GuideListener : Listener` for event handlers
  - Preserve all public API (100% backward compatible)

- **[Rebar.kt](path/to/Rebar.kt)**  
  - Update plugin event registration to use `RebarGuide.GuideListener`

## Impact
- ✅ **Breaking changes:** None
- ✅ **API compatibility:** All existing calls work unchanged
- ✅ **Fixes:** Kotlin addons can now use `RebarGuide` directly

## Verification
```kotlin
// Now compiles successfully from Kotlin addons:
val rootPage = RebarGuide.rootPage
RebarGuide.hideItem(key)
RebarGuide.open(player)
val page = RebarGuide.getOrCreateInfoPage(key)
```

## Type of Change
- [x] Bug fix (non-breaking)
- [ ] Feature
- [ ] Breaking change

## Testing
- [x] No compilation errors in modified files
- [x] IDE diagnostics clean (0 warnings/errors)
- [ ] Existing test suite passes
- [ ] Manual test with Kotlin addon project

## Related
- Issue: #[issue-number]
- Discussion: Kotlin compiler compatibility with complex companion objects

### DIFF
--- a/rebar/src/main/kotlin/io/github/pylonmc/rebar/Rebar.kt
+++ b/rebar/src/main/kotlin/io/github/pylonmc/rebar/Rebar.kt
@@ -135,7 +135,7 @@ object Rebar : JavaPlugin(), RebarAddon {
         pm.registerEvents(RebarDirectionalBlock, this)
         pm.registerEvents(FluidPipePlacementService, this)
         pm.registerEvents(RebarTickingBlock, this)
-        pm.registerEvents(RebarGuide, this)
+        pm.registerEvents(RebarGuide.GuideListener, this)
         pm.registerEvents(RebarLogisticBlock, this)
         pm.registerEvents(CargoRoutes, this)
         pm.registerEvents(CargoDuct, this)
--- a/rebar/src/main/kotlin/io/github/pylonmc/rebar/content/guide/RebarGuide.kt
+++ b/rebar/src/main/kotlin/io/github/pylonmc/rebar/content/guide/RebarGuide.kt
@@ -57,7 +57,7 @@ class RebarGuide(stack: ItemStack) : RebarItem(stack), RebarInteractor {
         }
     }
 
-    companion object : Listener {
+    companion object {
 
         @JvmField
         val KEY = rebarKey("guide")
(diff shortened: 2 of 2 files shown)


Seggan:
I am unable to reproduce the issue mentioned in the description with Rebar 0.39.1-26.1:
image
One issue is that you use a `getRootPage()` function instead of the `rootPage` property; the generated getters do not exist in Kotlin land, only the properties.

mc506lw:
Thanks. I'll fix the getRootPage() later (no PC right now). Just to note, the bottom screenshot shows hideItem being inaccessible – that's exactly what this PR fixes.

mc506lw:
Btw, just to be clear: this bug is Kotlin-only. The same code works fine in Java, but in Kotlin with compileOnly it just fails. That's why I ended up using reflection. Weird right?

Seggan:
I am aware. The screenshot I sent was in Kotlin. The issue does not happen for me.

Seggan:
Can I see your Gradle configuration?

mc506lw:
Did some more testing and found the pattern:
The local JAR from releases breaks, but the Maven remote one works. Same JAR (maybe), everything else is fine — only RebarGuide fails.

Is this a difference between the releases JAR and the repository version? How should we proceed with this PR?

Seggan:
Yeah I just tested it myself, using a local jar doesn't allow you to resolve anything at all. Just use the one from Maven Central, there's literally no reason not to. As for the PR itself, I think it can be closed.

mc506lw:
Okay, I also think closing it is the right way to go.

Issue #753, opened 2026-05-17
change item/block interface naming scheme to be more logical


LordIdra:
why the fuck is `RebarDispensable` called that

PR #786, opened 2026-05-28
Listener Interface Update

JustAHuman-xD:
#753 as well as renaming a number of the interface methods so that they aren't as generic, and fixing a number of mistakes in a few of the interfaces (i.e. the various projectile interfaces)

Currently this only covers items to see how everyone feels about this new scheme, if we like it Ill do the block & entity interfaces next

The idea is to split the concept of item interfaces & rebar item handlers, rebar item handlers only process events, and item interfaces add/change core functionality, any item interfaces that did both have been split into a designated handler and item interface instead

The diff isn't perfect because git is funny so some interfaces are shown to be deleted but have just been renamed (ex: RebarWeapon -> EntityAttackRebarItemHandler)


JustAHuman-xD:
Current before and after of item interfaces:

Before: 
image

After: 
image

Seggan:
you forgot to do `RebarArmorItem` -> `ArmorRebarItem`

rebar/src/main/kotlin/io/github/pylonmc/rebar/block/interfaces/InteractRebarBlockHandler.kt
Seggan, line 1: `InteractionRebarBlockHandler`

rebar/src/main/kotlin/io/github/pylonmc/rebar/entity/interfaces/DyeRebarEntityHandler.kt
Seggan, line 1: `DyeRebarEntityHandler`

rebar/src/main/kotlin/io/github/pylonmc/rebar/item/interfaces/UnmergeableRebarItem.kt
Seggan, line 1: `UmergeableRebarItem`

rebar/src/main/kotlin/io/github/pylonmc/rebar/entity/interfaces/DyeRebarEntityHandler.kt
JustAHuman-xD: might I ask why?

rebar/src/main/kotlin/io/github/pylonmc/rebar/block/interfaces/InteractRebarBlockHandler.kt
JustAHuman-xD: same question as for the comment below, all of the interact handlers are named like this

LordIdra:
The only thing id suggest is to remove the handlers folder and have it all in one directory, or organise them differently (e.g. by type like on the wiki). Just seems a bit weird to have some listeners at the top level and some in a subfolder, rather than all at the top level or all in subfolders. But it's a relatively minor thing

rebar/src/main/kotlin/io/github/pylonmc/rebar/block/interfaces/InteractRebarBlockHandler.kt
Seggan: Makes more sense to me

rebar/src/main/kotlin/io/github/pylonmc/rebar/entity/interfaces/DyeRebarEntityHandler.kt
Seggan: same as above

rebar/src/main/kotlin/io/github/pylonmc/rebar/block/interfaces/InteractRebarBlockHandler.kt
LordIdra: yeah it would be more consistent with the others tbh, like SneakRebarBlockHandler, JumpRebarBlockHandler, UnloadRebarBlockHandler

LordIdra: or like InteractRebarBlockHandler would be consistent

LordIdra:
Now that I'm looking at it again, does it make sense to split them up into Handler? Are we sure we don't want to just call them like SignRebarBlock, SneakRebarBlock, HopperRebarBlock, etc? Just so it's less wordy and for consistency (I don't think the 'handler' distinction is very useful)

JustAHuman-xD:
> Now that I'm looking at it again, does it make sense to split them up into Handler? Are we sure we don't want to just call them like SignRebarBlock, SneakRebarBlock, HopperRebarBlock, etc? Just so it's less wordy and for consistency (I don't think the 'handler' distinction is very useful)

I think so, the idea is that we strictly split up interfaces that only process events (handlers) with anything else. If you're searching for an inference to do something for the first time and you know that you're looking for a handler you won't waste time on all the other interfaces, and vice versa.

If we just call something "BedRebarBlock" instead of handler it could imply that that interface  *adds* the ability for a block to be used as a bed, because some of the interfaces (which are not handlers) add functionality that isn't there by default. (e.g. TickingRebarBlock for ex) By calling it a handler the implication is instead that you are only handling something that already happens, not making something happen.

JustAHuman-xD:
> The only thing id suggest is to remove the handlers folder and have it all in one directory, or organise them differently (e.g. by type like on the wiki). Just seems a bit weird to have some listeners at the top level and some in a subfolder, rather than all at the top level or all in subfolders. But it's a relatively minor thing

I mean its just organization, splitting handlers and non handlers, but like I could put them at the same level just different folders for ex

JustAHuman-xD:
todo: sort the provided item/block/entity interfaces (idra is doing this ty kind sir)

rebar/src/main/kotlin/io/github/pylonmc/rebar/nms/NmsAccessor.kt
```diff
@@ -86,6 +86,13 @@ interface NmsAccessor {
      */
     fun setFurnaceRecipeCache(block: Block, recipe: NamespacedKey)
 
+    /**
+     * Returns the weapon item for the entity
+     *
+     * For ex: If a player is spinning with a riptide trident, returns the trident, otherwise it may return the sword in the main hand
+     */
+    fun getWeaponItem(entity: Entity): ItemStack?
```
Seggan, line 96: Why can't this just be `itemInMainHand`?

JustAHuman-xD: Its listed in the javadoc but for ex: using riptide with a trident in your offhand

getWeaponItem is an nms method thats just not exposed to handle whatever slot has the active using weapon

Seggan: TIL you can use tridents in offhand

Seggan (approved):

LordIdra (approved):

PR #797, opened 2026-05-31
Fix BlockTextureEntity Brightness

### LINKED ISSUE #634: Block textures have wrong light level
**Issue reported by: overlordidra:**

JustAHuman-xD:
Fixes #634 

In order to give accurate lighting information we have to move the location of the entity to a block that has lighting data, specifically the one with the most light. of the rebar block itself is transparent then it will have lighting data (think glass) and we don't have to do anything, but if its opaque (think copper block) then we have to delegate to a neighbor that has the most light

that also means we have to watch for the neighbors changing

I removed the ability to change the transformation of the base BlockTextureEntity interface so that I can safely in the impl avoid sending unnecessary data.

In this I also fix a couple of bugs in the related systems (i.e. making a rebar block visible/culled every tick if it was within the always show radius or outside the cull radius instead of following the visible/hidden intervals)


Seggan (approved):
Understanding: 0%
Working: 100%

Approved ✔️

PR #807, opened 2026-06-06
Waila Suppliers

### LINKED ISSUE #700: Multiblocks that turn into phantom blocks do not remove WAILA override from components which leads to more errors

JustAHuman-xD:
adds waila suppliers so that we can have auto cleanup of proxy waila overrides, also adds RebarBlockPhantomEvent so that we can cleanup waila suppliers whenever a block turns into a phantom block (and anything else that needs to handle turning into phantom blocks can now use this event)

### DIFF
--- a/rebar/src/main/kotlin/io/github/pylonmc/rebar/waila/Waila.kt
+++ b/rebar/src/main/kotlin/io/github/pylonmc/rebar/waila/Waila.kt
@@ -356,13 +356,13 @@ class Waila private constructor(private val player: Player, playerConfig: Player
         }
 
         private fun removeOverrides(block: RebarBlock) {
-            blockOverrides.values.removeIf { supplier -> supplier is RebarBlockWailaSupplier && supplier.source == block }
-            entityOverrides.values.removeIf { supplier -> supplier is RebarBlockWailaSupplier && supplier.source == block }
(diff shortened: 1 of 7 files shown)


LordIdra (approved):

rebar/src/main/kotlin/io/github/pylonmc/rebar/waila/Waila.kt
```diff
@@ -311,8 +315,8 @@ class Waila private constructor(private val player: Player, playerConfig: Player
          * old override will be replaced.
          */
         @JvmStatic
-        fun addWailaOverride(entity: Entity, provider: (Player) -> WailaDisplay?) {
-            entityOverrides[entity.uniqueId] = provider
+        fun addWailaOverride(entity: Entity, supplier: WailaSupplier) {
+            entityOverrides[entity.uniqueId] = supplier
         }
```
LordIdra, line 389: I think these methods could have overloads for `(Player) -> WailaDisplay?` that construct RebarBlockWailaSupplier / RebarEntityWailaSupplier automatically? Just so it's nicer to call from Pylon/addons

LordIdra (changes requested):

Seggan (approved):

LordIdra (approved):

PR #818, opened 2026-06-08
WAILA refactor

### LINKED ISSUE #803: Waila Improvements
- block break progress w/ bossbar progress
- preferred tool prefix on waila *TOOL_ICON | WAILA"

### LINKED ISSUE #804: Higher-level WAILA API
Currently WAILA is in the format `Idra's Stupid Block | |||||||| 100% | Off`, i.e. vertical bars separating each section. It would make sense to abstract some WAILA functionality into a higher-level API so the above could be achieved with something like

```
WailaBuilder.of(getKey())
    .add(ProgressBarBuilder.recipeProgress(...))
    .add(Component.translateable("pylon.message." + isOn ? "on" : "off")
```

This would
- drastically simplify a lot of WAILA code
- reduce boilerplate
- remove some unecessary language file entries (including all/almost all WAILA language entires)
- make writing WAILAs a lot less cumbersome
- improve consistency across addons

WailaBuilder should implement ComponentLike, so you can still provide your own component as a WAILA and are not forced to use the WailaBuilder.

Related to https://github.com/pylonmc/rebar/issues/803

LordIdra:
Well this ended up being bigger than I expected
- Adds a tracker for block break progress which listens to `BlockBreakProgressUpdateEvent`
- Overhauls WAILA display construction to be built in segments with a `|` between each segment - this simplifies waila construction a LOT in some cases
- Adds a prefix to blocks which shows the required tool and whether the current tool in the player's hand can break the block
- Bossbar progress now indicates block break progress
- Removes default WAILA translation key as it is no longer needed
- Makes WAILA constructor private and introduces a number of static constructor methods including 
- Splits WAILA updates into target updates ('which block/entity is the player looking at' - ticks slowly) and contents updates ('what should the bossbar/action bar say for the block/entity the player is looking at' - ticks quickly) to mitigate the fucking ridiculous performance impact of doing 100s/1000s of raytraces per second while keeping WAILA contents updates snappy
- WAILA target is updated slower (factor of 4 by default) when player is not looking around (this is almost never noticeable in my testing and potentially quite a significant performance boost)
- Fixed children of translateable components not being updated
- Added `getPreferredTool` method (unfortunately needs hardcoded logic, not possible with nms either)

closes https://github.com/pylonmc/rebar/issues/803
closes https://github.com/pylonmc/rebar/issues/804


rebar/src/main/kotlin/io/github/pylonmc/rebar/block/BlockListener.kt
```diff
@@ -265,6 +270,15 @@ internal object BlockListener : MultiListener {
         }
     }
 
+    @EventHandler(priority = EventPriority.MONITOR, ignoreCancelled = true)
+    private fun onBlockBreakProgressUpdated(event: BlockBreakProgressUpdateEvent) {
```
Seggan, line 274: What about when player leaves

rebar/src/main/kotlin/io/github/pylonmc/rebar/waila/WailaDisplayBuilder.kt
```diff
@@ -0,0 +1,14 @@
+package io.github.pylonmc.rebar.waila
+
+import io.github.pylonmc.rebar.block.RebarBlock
+import io.github.pylonmc.rebar.item.RebarItem
+import net.kyori.adventure.text.Component
+
+/**
+ * A simple utility class for building WAILA displays.
+ */
+class WailaDisplayBuilder private constructor(text: Component) {
```
Seggan: This doesn't seem to actually do anything useful?

rebar/src/main/resources/lang/en.yml
```diff
@@ -15,20 +15,14 @@ item:
       <arrow> You can open the guide without this item with <insn>/rb</insn>
 
   phantom_block:
-    name: "<red>Phantom Block (%block%)"
```
Seggan, line 18: Why remove block param?

rebar/src/main/kotlin/io/github/pylonmc/rebar/block/BlockListener.kt
LordIdra: The break progress resets when this happens I'm pretty sure

rebar/src/main/kotlin/io/github/pylonmc/rebar/waila/WailaDisplayBuilder.kt
LordIdra: oops forgot to remove XD

rebar/src/main/resources/lang/en.yml
LordIdra: It's not removed, it's just added using the new WAILA API instead of in the lang file

rebar/src/main/kotlin/io/github/pylonmc/rebar/waila/Waila.kt
```diff
@@ -144,14 +164,26 @@ class Waila private constructor(private val player: Player, playerConfig: Player
...
-        rayTraceResult.hitEntity?.let { entity ->
+    private fun updateContents() {
+        if (targetEntity != null) {
             try {
-                var display = entityOverrides[entity.uniqueId]?.invoke(player)
+                val entity = Bukkit.getEntity(targetEntity!!)
+                if (entity == null || !entity.isValid) {
+                    targetEntity = null
+                    playerEyeLocationAtLastTargetUpdate = null // Force WAILA target to update
+                    updateTarget()
```
JustAHuman-xD: doesn't `updateTarget()` force WAILA target to update? Why are you assigning it to null as well

rebar/src/main/kotlin/io/github/pylonmc/rebar/waila/Waila.kt
```diff
@@ -177,15 +209,30 @@ class Waila private constructor(private val player: Player, playerConfig: Player
...
-        rayTraceResult.hitBlock?.let { block ->
+        } else if (targetBlock != null) {
             try {
-                var display = blockOverrides[block.position]?.invoke(player)
+                val block = targetBlock!!.block
+
+                if (block.isEmpty) {
+                    targetBlock = null
+                    playerEyeLocationAtLastTargetUpdate = null // Force WAILA target to update
+                    updateTarget()
```
JustAHuman-xD: Same thing here

rebar/src/main/kotlin/io/github/pylonmc/rebar/waila/Waila.kt
```diff
@@ -177,15 +209,30 @@ class Waila private constructor(private val player: Player, playerConfig: Player
...
+                    hide()
+                    return
+                }
+
+                var display = blockOverrides[targetBlock]?.invoke(player)
                     ?: block.let(BlockStorage::get)?.getWaila(player)
 
                 if (display == null && player.wailaConfig.vanillaWailaEnabled) {
```
JustAHuman-xD: We should probably update this check to also ensure that there is no rebar block at this block, otherwise in theory if a rebar block intentionally returned null for the display it would default to giving a vanilla WAILA for a rebar block

rebar/src/main/kotlin/io/github/pylonmc/rebar/waila/Waila.kt
```diff
@@ -233,14 +282,33 @@ class Waila private constructor(private val player: Player, playerConfig: Player
...
+                    // Delay for at most TARGET_TICK_INTERVAL * STATIONARY_TARGET_TICK_INTERVAL_MULTIPLIER ticks,
+                    // until the player moves their eyes
+                    for (i in 0..<RebarConfig.WailaConfig.STATIONARY_TARGET_TICK_INTERVAL_MULTIPLIER) {
+                        delayTicks(RebarConfig.WailaConfig.TARGET_TICK_INTERVAL.toLong())
+                        if (waila.playerEyeLocationAtLastTargetUpdate != player.eyeLocation) {
+                            break
+                        }
+                    }
+                }
+            }
```
JustAHuman-xD, line 309: More of a @Seggan question atp tbh, is it better to have these separate w/ the dedicated different delays, or would it be potentially better to have 1 job that delays at a common modulo of both delays (e.g. 1 tick delay) and then just checks if its an interval is met or not

rebar/src/main/resources/config.yml
```diff
@@ -42,11 +42,19 @@ confetti-creeper:
...
+
+  # The interval (in Minecraft ticks), between the WAILA display being updated for the player's target block/entity
+  # This can generally be kept low without any performance issues
+  contents-tick-interval: 2
 
   # The WAILA display types players can choose between
+  # Set to an empty list to disable WAILA entirely
   enabled-types:
     - actionbar
     - bossbar
```
JustAHuman-xD, line 80: we should have some option to enable waila for placeholder usage only

rebar/src/main/kotlin/io/github/pylonmc/rebar/waila/WailaDisplay.kt
```diff
@@ -1,17 +1,87 @@
...
+        @JvmStatic
+        fun of(component: Component) = WailaDisplay(component)
+
+        /**
+         * Constructs a new WAILA builder which begins with the name of the provided [block] and
+         * has progress equal to the break progress of the block.
+         */
+        @JvmStatic
+        fun of(block: RebarBlock, player: Player): WailaDisplay {
+            val name = Component.translatable("${block.key.namespace}.item.${block.key.key}.name")
```
JustAHuman-xD: what if a block doesn't have an item registered for it (I have at least one occurrence of this already in occult operations)

rebar/src/main/kotlin/io/github/pylonmc/rebar/waila/Waila.kt
```diff
@@ -233,14 +282,33 @@ class Waila private constructor(private val player: Player, playerConfig: Player
...
+                    delayTicks(RebarConfig.WailaConfig.CONTENTS_TICK_INTERVAL.toLong())
                 }
-            })
+            }
+
+            val updateTargetJob = Rebar.scope.launch {
+                delayTicks(1)
+                val waila = wailas[player.uniqueId]!!
+                while (true) {
+                    waila.updateTarget()
```
JustAHuman-xD, line 298: depending on the delays set in the config, updating the target & updating contents may not overlap nicely which could cause extra delay, should we not update the contents immediately after updating the target?

rebar/src/main/kotlin/io/github/pylonmc/rebar/util/RebarUtils.kt
```diff
@@ -747,4 +749,69 @@ fun ItemStack.overriddenDataTypes(): List<DataComponentType> {
     return dataTypes.filter { isDataOverridden(it) }
 }
 
-const val FLUID_EPSILON = 1.0e-6
\ No newline at end of file
+const val FLUID_EPSILON = 1.0e-6
+
+fun Material.getPreferredTool(): Material? {
```
JustAHuman-xD, line 766: still debating if we should pre-compute and/or cache this

rebar/src/main/kotlin/io/github/pylonmc/rebar/waila/Waila.kt
LordIdra: relic from old logic

rebar/src/main/kotlin/io/github/pylonmc/rebar/waila/Waila.kt
LordIdra: I don't see why that would be better? It would be a lot more convoluted in any case

rebar/src/main/resources/config.yml
LordIdra: I'll just add a `enable-waila: true` option atp

rebar/src/main/kotlin/io/github/pylonmc/rebar/waila/Waila.kt
LordIdra: yeah I can just update contents always when target updated

rebar/src/main/kotlin/io/github/pylonmc/rebar/util/RebarUtils.kt
LordIdra: I doubt it's going to matter, it's a really simple piece of code that does very little work (like 5 map lookups or something at most) and is only going to be called 10s of times a second

rebar/src/main/kotlin/io/github/pylonmc/rebar/waila/WailaDisplay.kt
LordIdra: added an overload to allow name to be manually supplied

rebar/src/main/kotlin/io/github/pylonmc/rebar/waila/WailaDisplay.kt
```diff
@@ -1,17 +1,112 @@
...
+         * If the block has a preferred tool, the WAILA starts with said tool and a + or - depending on
+         * whether the block will drop when broken or not.
+         *
+         * The WAILA will then contain the given [name].
+         *
+         * The returned WAILA has progress equal to the break progress of the block.
+         */
+        @JvmStatic
+        fun of(block: RebarBlock, player: Player, name: Component): WailaDisplay {
+            val name = Component.translatable("${block.key.namespace}.item.${block.key.key}.name")
```
JustAHuman-xD: You never removed the original name declaration so you hide the parameter with this lol

JustAHuman-xD (changes requested):

rebar/src/main/kotlin/io/github/pylonmc/rebar/waila/WailaDisplay.kt
LordIdra: oh oops lmao

JustAHuman-xD (approved):

PR #827, opened 2026-06-08
Batch Recipe (Un)Registration

### LINKED ISSUE #822: Improve recipe loading times
why are they taking 2 seconds

JustAHuman-xD:
Bukkit will completely reload the NMS RecipeManager for every single recipe that is (un)registered, so when registering lots of recipes at once using the Bukkit methods, it does a LOT of unnecessary reloading, which gets exponentially slower with each recipe (un)registered. (This gets even worse when there are players online at the time of registration, though currently that does not apply to us as this happens while the server is starting & we do not reload recipes ever, however if we did it would be EVEN better as previously it would have to reload once per recipe to unregister and then once per recipe to re-register)

We can solve this problem by batching the recipes & registering them ourselves using NMS instead of going through Bukkit, allowing us to register/unregister anything necessary and only reload the RecipeManager one time.

Difference in recipe load time on my machine:
- Not Debugging: `0.611s` -> `0.219s`
- Debugging: `0.805s` -> `0.273s`

*I think that machines which have a longer recipe load time before batching will benefit even more proportionately than I do, but I need other people's numbers to back that up*

I also took some spark reports (not debugging) of only the recipe loading:
Before batching: https://spark.lucko.me/LPtzjd1uLt
After batching: https://spark.lucko.me/9n3gWagj8A

image

### DIFF
--- a/nms/src/main/kotlin/io/github/pylonmc/rebar/nms/NmsAccessorImpl.kt
+++ b/nms/src/main/kotlin/io/github/pylonmc/rebar/nms/NmsAccessorImpl.kt
@@ -8,17 +8,19 @@ import io.github.pylonmc.rebar.async.PlayerScope
 import io.github.pylonmc.rebar.block.RebarBlock
 import io.github.pylonmc.rebar.entity.packet.BlockTextureEntity
 import io.github.pylonmc.rebar.i18n.PlayerTranslationHandler
-import io.github.pylonmc.rebar.nms.packet.PlayerPacketHandler
 import io.github.pylonmc.rebar.item.ItemTypeWrapper
 import io.github.pylonmc.rebar.nms.entity.BlockTextureEntityImpl
 import io.github.pylonmc.rebar.nms.inventory.KeyedContainerListener
+import io.github.pylonmc.rebar.nms.packet.PlayerPacketHandler
 import io.github.pylonmc.rebar.nms.recipe.AccessibleCachedCheck
 import io.github.pylonmc.rebar.nms.recipe.HandlerRecipeBookClick
+import io.github.pylonmc.rebar.nms.recipe.RecipeMapper
 import io.github.pylonmc.rebar.util.position.BlockPosition
 import io.papermc.paper.adventure.PaperAdventure
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
 import net.kyori.adventure.text.Component
+import net.minecraft.advancements.AdvancementRewards.Builder.recipe
 import net.minecraft.commands.arguments.item.ItemParser
 import net.minecraft.core.BlockPos
 import net.minecraft.core.registries.Registries
@@ -35,6 +37,7 @@ import net.minecraft.world.item.crafting.RecipeManager
 import net.minecraft.world.level.block.entity.AbstractFurnaceBlockEntity
 import net.minecraft.world.level.block.state.properties.Property
 import net.minecraft.world.phys.BlockHitResult
+import org.bukkit.Bukkit
 import org.bukkit.Material
 import org.bukkit.NamespacedKey
 import org.bukkit.World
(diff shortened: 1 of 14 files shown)


LordIdra:
Average 1.7-2.2 seconds to ~0.4 seconds on my machine

It's annoying that it requires this much code but it's all pretty simple. I think it's probably worth adding because while it's not terrible rn, the time spent is only going to go up and addons will increase it further, it could quite easily reach 10+ seconds on non-god-tier hardware especially if an addon add lots of vanilla recipes

Seggan:
Agree, we should keep this. Though maybe a Paper PR is warranted?

JustAHuman-xD:
> Agree, we should keep this. Though maybe a Paper PR is warranted?

Yeah I already added it to my list of things to PR to paper

balugaq:
QQ_1780941312878

I noticed that `enabledFlags` can control `finalizeRecipeLoading`. Why not consider using Reflection to change the value of `enabledFlags` to skip `finalizeRecipeLoading`, thereby achieving a speedup?

JustAHuman-xD:
> I noticed that `enabledFlags` can control `finalizeRecipeLoading`. Why not consider using Reflection to change the value of `enabledFlags` to skip `finalizeRecipeLoading`, thereby achieving a speedup?

If we skip `finalizeRecipeLoading` we would break the recipe manager? As well as desync every single client connected to the server. lol

glomdom:
Ran this on my machine and got:
master branch: Loaded recipes in 1.25s
this PR's branch: Loaded recipes in 0.47s

balugaq:
> > I noticed that `enabledFlags` can control `finalizeRecipeLoading`. Why not consider using Reflection to change the value of `enabledFlags` to skip `finalizeRecipeLoading`, thereby achieving a speedup?
> 
> If we skip `finalizeRecipeLoading` we would break the recipe manager? As well as desync every single client connected to the server. lol

Because `Bukkit.addRecipe()` is required to be executed on the main thread, as long as we can ensure that all loading is completed within a certain period of time and then reset `enabledFlags` back to its original value, it won't affect normal players' usage.

Moreover, enabledFlags is not used in many places — it is ultimately only accessed in `RecipeManager.IngredientCollector#asPropertySet` when `RecipeManager#filterDisabled` is called.

JustAHuman-xD:
> Because `Bukkit.addRecipe()` is required to be executed on the main thread, as long as we can ensure that all loading is completed within a certain period of time and then reset `enabledFlags` back to its original value, it won't affect normal players' usage.
> 
> Moreover, enabledFlags is not used in many places — it is ultimately only accessed in `RecipeManager.IngredientCollector#asPropertySet` when `RecipeManager#filterDisabled` is called.

*The whole point of this PR is to not call Bukkit.addRecipe*

balugaq:
> _The whole point of this PR is to not call Bukkit.addRecipe_

Isn't it about speed up?

JustAHuman-xD:
> Isn't it about speed up?

*yes, which in order to do, we stop using Bukkit#addRecipe*

JustAHuman-xD:
If the referenced paper PR gets merged all of the nms additions in this can be nuked and we can switch to the methods added in it

balugaq:
Made a simple benchmark (on my machine), let reloading Pylon count as one reload.

① = Normal MSUA
② = Normal Rebar
③ = Accelerated MSUA (Reflection modified `enabledFlags`)
④ = Accelerated Rebar (This PR)

| Item | Spark| Log | Reload Count | enablePlugin Total (ms) | enablePlugin Avg (ms) | loadFromConfig Total (ms) | loadFromConfig Avg (ms) |
|------|-----|------------|--------------|-------------------------|-----------------------|---------------------------|-------------------------|
| ① + ② | [Link](https://spark.lucko.me/wyuMm3PPDJ) | [Log](https://mclo.gs/OR2CT7H) | 8 | 74248 | 9281.00 | 65088 | 8136.00 |
| ① + ③ | [Link](https://spark.lucko.me/zV5shU0mm8) | [Log](https://mclo.gs/7pYhf3Q) | 49 | 77552 | 1582.69 | 47324 | 965.80 |
| ④ + ② | [Link](https://spark.lucko.me/mbItn2Ock0) | [Log](https://mclo.gs/m2PVrEe) | 30 | 62064 | 2068.80 | 18476 | 615.87 |

**MSUA is an addon that supports Rebar addon to be reloaded.*

Based on the data, this PR indeed provides a significant speed improvement, and it is even faster than using Reflection.

Intybyte:
Loading recipes master: 1.588s, 3.511s, 1.163s
Loading recipes justin nms edition ™️: 0.389s, 0.41s, 0.369s

Seggan (approved):

JustAHuman-xD:
this also optimizes a number of other things that happen to occur during recipe loading & registration

JustAHuman-xD:
Also with the latest commit the time for me is now 0.1s +- 0.01s

PR #828, opened 2026-06-08
ItemStackBuilder#copyOf & ItemStackBuilder#asOne

JustAHuman-xD:


LordIdra (approved):
I think 'single' or 'one' would be better than 'singleton' but I don't think it matters much

JustAHuman-xD:
> I think 'single' or 'one' would be better than 'singleton' but I don't think it matters much

Would you like 'asOne', to mirror the item stack method.

LordIdra:
> > I think 'single' or 'one' would be better than 'singleton' but I don't think it matters much
> 
> Would you like 'asOne', to mirror the item stack method.

Oh yeah that's fine

Issue #792, opened 2026-05-30
Line break does not account for languages without spaces


OhmV-IR:
**Issue reported by: lwj_666:**
[View original discord thread](https://discord.com/channels/1329177304857055273/1492744568746016880)
Maybe <story> has word number limit?
i add 1 line break (line 1731) and it backs to normal

![Screenshot_20260412_122809.jpg](https://cdn.discordapp.com/attachments/1492744568746016880/1492744569056661504/Screenshot_20260412_122809.jpg?ex=6a1bba37&is=6a1a68b7&hm=ffdccc0e68e0b9feaefa1deb4dfb9554c40c8830f791415eb5a7f00922c6698b&)
![Image_1775968321566.png](https://cdn.discordapp.com/attachments/1492744568746016880/1492744569820020858/Image_1775968321566.png?ex=6a1bba37&is=6a1a68b7&hm=81f644f696bf1969b85aa576e3ed1828cf00abcbf5cc98ae70e28f2eac354646&)
![Screenshot_20260412_122940.jpg](https://cdn.discordapp.com/attachments/1492744568746016880/1492744570255970334/Screenshot_20260412_122940.jpg?ex=6a1bba37&is=6a1a68b7&hm=ede75697402f0b3a242e9229f2b3757bce1f53d03f338fe33d75874137afa2da&)
![Image_1775968269051.png](https://cdn.discordapp.com/attachments/1492744568746016880/1492744570675527740/Image_1775968269051.png?ex=6a1bba37&is=6a1a68b7&hm=f8769f1caac27fd2fac83048f30ed10704ebb17ee336ce750f22264ac4c9f2a7&)

OhmV-IR:
**Message from schn:**
It looks like the auto line break doesn't account for CJK

OhmV-IR:
**Message from lwj_666:**
wait, we have auto line break?

OhmV-IR:
**Message from lwj_666:**
why the pillar's story has line breaks..

![Screenshot_20260412_174152.jpg](https://cdn.discordapp.com/attachments/1492744568746016880/1492822309998690394/Screenshot_20260412_174152.jpg?ex=6a1c029e&is=6a1ab11e&hm=9cd4499409adf80e518894e74192476458ad53fcad424a4a4ced5534dfcdaaae&)

OhmV-IR:
**Message from seggan:**
No idea

PR #831, opened 2026-06-08
Fix CJK line wrapping (#792) and harden FluidIntersectionMarker WAILA

KevinWoodWL:
## What

Four small, self-contained robustness fixes that prevent server-side crashes and log spam:

### 1. `PlayerPacketHandler` — don't mutate shared `ItemStack`s on the Netty thread
Outgoing item translation previously modified the NMS `ItemStack`s in place. This handler runs on Netty's I/O thread, while those same stacks can be touched by the main thread, so mutating their component maps races. On 1.21.6+ this surfaces as `ArrayIndexOutOfBoundsException` inside fastutil's `Reference2ObjectArrayMap` and then `Can't find id for 'null'` when the corrupted packet is encoded.

Now we translate **copies** and rebuild the affected packets (`ContainerSetContent`, `ContainerSetSlot`, `SetCursorItem`, `SetEquipment`). Packet rewriting is also wrapped in try/catch so a failure falls back to sending the original packet instead of dropping the connection.

### 2. `LineWrapping` — infinite loop / index out of bounds on spaceless text
The word-wrap loop assumed text contained spaces and that no single token exceeded the wrap limit. With CJK text (no spaces) or an over-long token it could loop forever or throw `StringIndexOutOfBoundsException`. Added guards for empty content, non-positive remaining width, and clamps the cut index to the content length.

### 3. `FluidButton` — `IndexOutOfBounds` on empty fluid list
`currentFluid` indexed into the fluids list unconditionally. Made it nullable via `getOrNull`, fall back to a shared error item, and bail out of click handling when no fluid is resolvable.

### 4. `FluidIntersectionMarker` — NPE in WAILA when pipe display is missing
If the held intersection display or its connected pipe display entity has been lost (e.g. entity cleanup, world data issues), the `pipe` getter NPE'd inside the WAILA update tick. Added `pipeOrNull()`; `getWaila` now hides the WAILA and `getPickItem` returns null instead of throwing.

## Why

All four were hit on a live server and either crashed a subsystem or spammed the console every tick.

## Notes

Each fix is independent and touches a single file — happy to split into separate PRs if preferred.

### DIFF
--- a/rebar/src/main/kotlin/io/github/pylonmc/rebar/content/fluid/FluidIntersectionMarker.kt
+++ b/rebar/src/main/kotlin/io/github/pylonmc/rebar/content/fluid/FluidIntersectionMarker.kt
@@ -49,19 +49,23 @@ class FluidIntersectionMarker : RebarBlock, EntityHolderRebarBlock, BlockBreakRe
         }
     }
 
-    override fun getWaila(player: Player): WailaDisplay?
-        = WailaDisplay(defaultWailaTranslationKey.arguments(RebarArgument.of("pipe", this.pipe.stack.effectiveName())))
+    override fun getWaila(player: Player): WailaDisplay? {
+        val pipeItem = pipeOrNull() ?: return null
+        return WailaDisplay(defaultWailaTranslationKey.arguments(RebarArgument.of("pipe", pipeItem.stack.effectiveName())))
+    }
 
     val pipe: RebarItem
-        get() {
-            check(fluidIntersectionDisplay.connectedPipeDisplays.isNotEmpty())
-            val uuid = fluidIntersectionDisplay.connectedPipeDisplays.iterator().next()
-            return EntityStorage.getAs<FluidPipeDisplay?>(uuid)!!.pipe
-        }
+        get() = pipeOrNull() ?: error("FluidIntersectionMarker has no resolvable pipe display")
+
+    private fun pipeOrNull(): RebarItem? {
+        val display = getHeldRebarEntity(FluidIntersectionDisplay::class.java, "intersection") ?: return null
+        val uuid = display.connectedPipeDisplays.firstOrNull() ?: return null
+        return EntityStorage.getAs<FluidPipeDisplay?>(uuid)?.pipe
+    }
 
     override fun getDropItem(context: BlockBreakContext) = null
 
-    override fun getPickItem(player: Player) = pipe.stack
+    override fun getPickItem(player: Player) = pipeOrNull()?.stack
 
     companion object {
         @JvmField
--- a/rebar/src/main/kotlin/io/github/pylonmc/rebar/i18n/LineWrapping.kt
+++ b/rebar/src/main/kotlin/io/github/pylonmc/rebar/i18n/LineWrapping.kt
@@ -90,18 +90,44 @@ private fun wrapLine(
         // which are handled clientside) but this is fine, it can't be perfect.
         var content = component.content()
         while (currentLineLength + content.length > RebarConfig.TRANSLATION_WRAP_LIMIT) {
+            // How many more characters fit on the current line
+            val remaining = RebarConfig.TRANSLATION_WRAP_LIMIT - currentLineLength
 
-            // Make sure we snap to the end of a word (i.e. don't cut words in half)
-            var endIndex = RebarConfig.TRANSLATION_WRAP_LIMIT - currentLineLength
-            while (endIndex != 0 && content[endIndex] != ' ') {
+            // Current line is already full — flush it and restart on a fresh line
+            if (remaining <= 0) {
+                lines.add(currentLine)
+                currentLine = DEFAULT_COMPONENT
+                currentLineLength = 0
+                continue
+            }
+
+            // Try to snap to the end of a word so we don't cut words in half.
+            // (content.length > remaining here, so content[endIndex] is always in bounds.)
+            var endIndex = remaining
+            while (endIndex > 0 && content[endIndex] != ' ') {
                 endIndex -= 1
             }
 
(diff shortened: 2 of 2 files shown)


rebar/src/main/kotlin/io/github/pylonmc/rebar/content/fluid/FluidIntersectionMarker.kt
```diff
@@ -49,19 +49,27 @@ class FluidIntersectionMarker : RebarBlock, EntityHolderRebarBlock, BlockBreakRe
...
-            val uuid = fluidIntersectionDisplay.connectedPipeDisplays.iterator().next()
-            return EntityStorage.getAs<FluidPipeDisplay?>(uuid)!!.pipe
+        get() = pipeOrNull() ?: error("FluidIntersectionMarker has no resolvable pipe display")
+
+    private fun pipeOrNull(): RebarItem? {
+        val display = try {
+            fluidIntersectionDisplay
+        } catch (_: Throwable) {
+            return null
         }
```
LordIdra: Use `getHeldRebarEntity(FluidIntersectionDisplay::class.java, "intersection")` instead of try-catching

rebar/src/main/kotlin/io/github/pylonmc/rebar/guide/button/FluidButton.kt
LordIdra, line 1: The changes in this file aren't really necessary, as the FluidButton constructor expects at least 1 fluid to be provided and this is explicitly stated in documentation. I would just add an initialiser which checks that at least one fluid has been provided and throws an exception if not, to provide a clearer error message. But there isn't really a need for a player-facing error message here

rebar/src/main/kotlin/io/github/pylonmc/rebar/i18n/LineWrapping.kt
```diff
@@ -90,18 +90,33 @@ private fun wrapLine(
...
+            if (endIndex <= 0) {
+                endIndex = maxCharsOnLine.coerceAtMost(content.length)
+            }
+
             currentLine = currentLine.append(Component.text(content.substring(0, endIndex)).style(style))
             lines.add(currentLine)
             currentLine = DEFAULT_COMPONENT
             currentLineLength = 0
-            content = if (endIndex+1 == content.length) "" else content.substring(endIndex+1)
+            content = if (endIndex >= content.length) "" else content.substring(endIndex + 1)
```
LordIdra: What is this change for? The new code is not semantically equivalent

rebar/src/main/kotlin/io/github/pylonmc/rebar/i18n/LineWrapping.kt
```diff
@@ -90,18 +90,33 @@ private fun wrapLine(
         // which are handled clientside) but this is fine, it can't be perfect.
         var content = component.content()
         while (currentLineLength + content.length > RebarConfig.TRANSLATION_WRAP_LIMIT) {
+            if (content.isEmpty()) {
+                break
+            }
 
             // Make sure we snap to the end of a word (i.e. don't cut words in half)
```
LordIdra: Forgot to move this comment down too

rebar/src/main/kotlin/io/github/pylonmc/rebar/i18n/LineWrapping.kt
```diff
@@ -90,18 +90,33 @@ private fun wrapLine(
...
-            var endIndex = RebarConfig.TRANSLATION_WRAP_LIMIT - currentLineLength
-            while (endIndex != 0 && content[endIndex] != ' ') {
+            val maxCharsOnLine = RebarConfig.TRANSLATION_WRAP_LIMIT - currentLineLength
+            if (maxCharsOnLine <= 0) {
+                lines.add(currentLine)
+                currentLine = DEFAULT_COMPONENT
+                currentLineLength = 0
+                continue
+            }
```
LordIdra: I would add a comment to explain why this is necessary

LordIdra (changes requested):
Could use @JustAHuman-xD eyes on packet handler stuff

Does this close https://github.com/pylonmc/rebar/issues/792? From what I gather it doesn't actually implement line wrapping for strings without spaces, just stops them from erroring?

Seggan:
Having written the original packet handler, I can tell you that we mutate the internal stacks for a reason. This is extremely performance sensitive code, and even doing something like cloning an item stack has severe performance impacts, which is why me mutate the internal stack. Those we mutate are already cloned by the packet system, so I find it very hard to believe that a race condition can even happen here given that this is in the middle of sending the packet over the network. And yes all my assetions I have tested.

LordIdra:
ah thanks

KevinWoodWL:
Fair enough, thanks for explaining! Dropped the packet handler change. The crash I hit had invui's PacketListener in the stack too, so it's probably an interaction with that rather than your code, I'll look into it separately with a proper repro. Also reverted the FluidButton change since the init already covers it. PR is now just the CJK line-wrap fix (#792) and the FluidIntersectionMarker WAILA null-safety (switched to getHeldRebarEntity as @LordIdra suggested).

rebar/src/main/kotlin/io/github/pylonmc/rebar/guide/button/FluidButton.kt
JustAHuman-xD: actually this issue is handled in #830

LordIdra:
To re-iterate: Does this close https://github.com/pylonmc/rebar/issues/792? From what I gather it doesn't actually implement line wrapping for strings without spaces, just stops them from erroring?

LordIdra:
Closing as stale; opened [an issue](https://github.com/pylonmc/rebar/issues/865) for the pipe WAILA thing

PR #838, opened 2026-06-11
Entity Storage Fixes

JustAHuman-xD:
Currently EntityStorage saves the entities on entity removal, however that behavior is undefined and there is no guarantee that those changes will be persisted, I've changed it to entity unload instead. I also renamed the RebarEntityDeathEvent to be RebarEntityRemoveEvent as the name was misleading, death is not the same as removal, removal can be caused by death, but it isn't implicitly bound to it.

Ive also updated the death handler to listen to the real death event, not ours
1. because its been renamed and like I said isnt just death
2. because our event is called on the monitor priority of the root event, so in effect all handlers of our event are forced into monitor priority, so by switching to the actual event and not our own, implementing classes can still listen at any priority

For those reasons the new handler i added for actually handling entity removal, also uses the real event, not ours

### DIFF
--- a/rebar/src/main/kotlin/io/github/pylonmc/rebar/Rebar.kt
+++ b/rebar/src/main/kotlin/io/github/pylonmc/rebar/Rebar.kt
@@ -239,6 +239,7 @@ object Rebar : JavaPlugin(), RebarAddon {
         PathfindRebarEntityHandler.register(this, pm)
         PiglinRebarEntityHandler.register(this, pm)
         ProjectileRebarEntityHandler.register(this, pm)
+        RemoveRebarEntityHandler.register(this, pm)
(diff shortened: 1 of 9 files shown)


LordIdra:
@JustAHuman-xD Just to confirm, have you tested that the new changes work? The old implementation was in the state it was because the paper API was rather inconsistent wrt death/unload events but it has been a while since then, is that no longer the case as far as you can tell?

JustAHuman-xD:
> @JustAHuman-xD Just to confirm, have you tested that the new changes work? The old implementation was in the state it was because the paper API was rather inconsistent wrt death/unload events but it has been a while since then, is that no longer the case as far as you can tell?

Yeah I was using this on my branch with the RedstoneLink which was using the ghost block entities.

LordIdra (approved):

PR #843, opened 2026-06-14
Remove registry keys

### LINKED ISSUE #842: Remove registry keys
not used anywhere, and probably redundant as any addon using another addon's registry would softdepend on said addon anyway (or probably should). would be a small but nice simplification

LordIdra:
closes https://github.com/pylonmc/rebar/issues/842


rebar/src/main/kotlin/io/github/pylonmc/rebar/command/RegistryCommandArgument.kt
```diff
@@ -18,7 +18,7 @@ class RegistryCommandArgument<T : Keyed>(private val registry: RebarRegistry<T>)
 
     @Suppress("PrivatePropertyName")
     private val ERROR_UNKNOWN = DynamicCommandExceptionType {
-        MessageComponentSerializer.message().serialize(Component.text("Unknown key in ${registry.key}: $it"))
+        MessageComponentSerializer.message().serialize(Component.text("Unknown key: $it"))
```
JustAHuman-xD, line 21: I will say this is an unfortunate loss

Seggan (approved):
It is indeed an unfortunate loss

PR #845, opened 2026-06-14
Separate story text from lore and allow disabling story text

LordIdra:
- Separates story text from lore by adding a `story` key for items (see https://github.com/pylonmc/pylon/pull/1065)
- Updates translator to only add story text if player has it enabled and locale has the story text available
- Moves guide player settings into main player settings page object (I think this makes sense but not sure, any other input?)


LordIdra:
closes https://github.com/pylonmc/rebar/issues/790

JustAHuman-xD (approved):
looks good but untested

PR #846, opened 2026-06-14
Misc tweaks

### LINKED ISSUE #677: Make RebarGuiBlock more flexible
Currently after it creates the gui for the first time it gets stored in memory and there is no way to modify it or to overwrite the one stored in memory.

### LINKED ISSUE #798: /rb reloadlang

### LINKED ISSUE #823: Remove bypass permission for researches
not used any more

### LINKED ISSUE #824: Refuse to start on version mismatch
The current behaviour is a warning in console but this is very easy to miss, so we should just throw an exception and prevent Rebar from loading. There can be a config option to bypass this.

### LINKED ISSUE #825: Remove 'Hints' underline and change hints to be italic
The 'hints' heading made sense when we had lots of hints but it just looks ugly now. Instead, I think hints should be in italics to differentiate them (and possibly use a different color)

### LINKED ISSUE #844: Automatically detect supported languages from `lang` folder instead of requiring addons to provide a list

LordIdra:
closes https://github.com/pylonmc/rebar/issues/823
closes https://github.com/pylonmc/rebar/issues/824
closes https://github.com/pylonmc/rebar/issues/825
closes https://github.com/pylonmc/rebar/issues/677
closes https://github.com/pylonmc/rebar/issues/844
closes https://github.com/pylonmc/rebar/issues/798

### DIFF
--- a/rebar/src/main/kotlin/io/github/pylonmc/rebar/block/interfaces/GuiRebarBlock.kt
+++ b/rebar/src/main/kotlin/io/github/pylonmc/rebar/block/interfaces/GuiRebarBlock.kt
@@ -7,6 +7,7 @@ import io.github.pylonmc.rebar.event.RebarBlockLoadEvent
 import io.github.pylonmc.rebar.event.RebarBlockPlaceEvent
 import io.github.pylonmc.rebar.event.RebarBlockUnloadEvent
 import net.kyori.adventure.text.Component
+import org.bukkit.entity.Player
 import org.bukkit.event.Event
 import org.bukkit.event.EventHandler
 import org.bukkit.event.EventPriority
@@ -32,15 +33,43 @@ import java.util.IdentityHashMap
 interface GuiRebarBlock : NoVanillaInventoryRebarBlock {
 
     /**
+     * The title of the GUI
+     */
+    val guiTitle: Component
+        get() = (this as RebarBlock).nameTranslationKey
+
+    /**
      * Returns the block's GUI. Called when a block is created.
      */
     fun createGui(): Gui
 
     /**
-     * The title of the GUI
+     * Refreshes the stores GUI by calling [createGui] again.
+     *
+     * If players have the GUI already open, it will be closed and then re-opened with the
+     * new GUI manually.
+     *
+     * Strongly consider [updating individual items](https://docs.xenondevs.xyz/invui/item/)
+     * before you use this method, to prevent having to constantly close and re-open windows.
      */
-    val guiTitle: Component
-        get() = (this as RebarBlock).nameTranslationKey
+    fun refreshGui() {
+        val oldGui = guiBlocks[this]
+        val players = oldGui?.windows?.map { it.viewer } ?: emptyList()
+        oldGui?.windows?.forEach { it.close() }
+        guiBlocks[this] = createGui()
+        for (player in players) {
+            open(player)
+        }
+    }
+
+    fun open(player: Player) {
+            Window.builder()
+                .setUpperGui(guiBlocks[this]!!)
+                .setTitle(guiTitle)
+                .setViewer(player)
+                .build()
+                .open()
+    }
 
     companion object : Listener {
         private val guiBlocks = IdentityHashMap<GuiRebarBlock, Gui>()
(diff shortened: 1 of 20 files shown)


JustAHuman-xD (approved):
lgtm, untested

rebar/src/main/kotlin/io/github/pylonmc/rebar/block/interfaces/GuiRebarBlock.kt
```diff
@@ -31,16 +32,44 @@ import java.util.IdentityHashMap
...
      */
-    val guiTitle: Component
-        get() = (this as RebarBlock).nameTranslationKey
+    fun refreshGui() {
+        val oldGui = guiBlocks[this]
+        val players = oldGui?.windows?.map { it.viewer } ?: emptyList()
+        oldGui?.windows?.forEach { it.close() }
+        guiBlocks[this] = createGui()
+        for (player in players) {
+            open(player)
```
Seggan, line 61: This doesn't work if the player is in a sub menu or smth

rebar/src/main/kotlin/io/github/pylonmc/rebar/i18n/RebarTranslator.kt
```diff
@@ -52,32 +54,46 @@ class RebarTranslator private constructor(private val addon: RebarAddon) : Trans
...
                     split.getOrNull(2).orEmpty()
-                ) to ConfigSection.fromOrThrow(it)
+                )
+                val config = ConfigSection.fromOrThrow(it)
+                translations[locale] = config
             }
         }
     }
 
+    fun reload() {
```
Seggan, line 92: Need to resend stuff to players

test/src/main/java/io/github/pylonmc/rebar/test/RebarTest.java
```diff
@@ -212,4 +207,9 @@ public void onEnable() {
     public @NotNull Material getMaterial() {
         return Material.WAXED_WEATHERED_CUT_COPPER_STAIRS;
     }
+
+    @Override
+    public @NotNull Locale getDefaultLanguage() {
+        return Locale.of("none");
```
Seggan, line 213: Is this even valid?

rebar/src/main/kotlin/io/github/pylonmc/rebar/block/interfaces/GuiRebarBlock.kt
LordIdra: No way around that unfortunately; docs discourage using this method where possible

rebar/src/main/kotlin/io/github/pylonmc/rebar/i18n/RebarTranslator.kt
LordIdra: What do you mean?

test/src/main/java/io/github/pylonmc/rebar/test/RebarTest.java
LordIdra: No but rebar test doesn't have a language file anyway and the tests run fine with it, we can change it if we add a language file for whatever reason

rebar/src/main/kotlin/io/github/pylonmc/rebar/i18n/RebarTranslator.kt
Seggan: If you reload lang the players won't get the updated lang until something forces the item to get resent.

LordIdra: Oh I see. I don't think we can really do this though? Or do you know a way?

Seggan: We have a whole method, it's called resendInventory iirc

JustAHuman-xD: Yeah its an nms method

LordIdra: This results in IndexOutOfBounds exceptions when any invui GUI is open. I will open an issue. It works fine for regular inventories though

Seggan (approved):

PR #851, opened 2026-06-15
Fix ticking entity startup and async texture lighting

KevinWoodWL:
﻿## Summary
- Ensure deserialized ticking entities are present in the ticking map before assigning their coroutine job, so later entities sharing a schema still tick.
- Avoid reading neighbor block states for texture lighting when the neighbor chunk is not loaded, preventing async chunk-load deadlocks during culling.

## Testing
- `./gradlew.bat :rebar:compileKotlin :nms:compileKotlin`

### DIFF
--- a/nms/src/main/kotlin/io/github/pylonmc/rebar/nms/entity/BlockTextureEntityImpl.kt
+++ b/nms/src/main/kotlin/io/github/pylonmc/rebar/nms/entity/BlockTextureEntityImpl.kt
@@ -365,6 +365,9 @@ class BlockTextureEntityImpl : BlockTextureEntity, SyncedDataHolder {
         return stateUpdated
     }
 
+    private fun CraftBlock.isChunkLoadedSafe(): Boolean =
+        world.isChunkLoaded(x shr 4, z shr 4)
+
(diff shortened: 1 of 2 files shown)


LordIdra:
Hi, thanks for the PR:
1. The ticking logic issue has been addressed already in https://github.com/pylonmc/rebar/pull/849 with a simpler fix. (The code in this PR re-implements the logic already run by calling `tickingBlock.tickingData`)
2. The lighting logic bug is definitely an issue, although from talking with @JustAHuman-xD  it would only occur on a very specific edge condition and would very likely result in a simple error rather than a deadlock
3. Please do not submit AI generated pull requests, which this clearly is, given the nature of the changes, the structure of the PR description, and the fact that you removed [codex] from the title. If we want AI-generated pull requests (which we do not), we can very easily generate them ourselves with less effort. This is just a waste of our time so please do not do it.

@JustAHuman-xD has said he'll apply a simpler fix to the lighting logic chunk unloading issue in a separate PR. If you find issues in the future, please open an issue or write a patch by hand instead of using AI and we are more than happy to address that and help you improve the patch as long as we are talking to a human.

PR #854, opened 2026-06-16
Fix type casting for inventoryView in addSlotChangedListener

TagDL:
When open portable dustbin or portable ender chest it will cause ClassCastException, so I just let it return null if cast class fail


LordIdra (approved):
Nice catch lmao

PR #862, opened 2026-06-29
Various Utils

JustAHuman-xD:
Needed for a pylon pr about to be opened


rebar/src/main/kotlin/io/github/pylonmc/rebar/item/builder/CustomModelDataBuilder.kt
```diff
@@ -0,0 +1,86 @@
+package io.github.pylonmc.rebar.item.builder
+
+import io.papermc.paper.datacomponent.DataComponentTypes
+import io.papermc.paper.datacomponent.item.CustomModelData
+import org.bukkit.Color
+import org.bukkit.inventory.ItemStack
+
+@Suppress("UnstableApiUsage")
+class CustomModelDataBuilder {
```
LordIdra, line 15: Might be worth documenting why this exists (immutability of CustomModelData) since it's a bit confusing if you don't know the context as this has all the same methods as CustomModelData.Builder

rebar/src/main/kotlin/io/github/pylonmc/rebar/block/RebarBlock.kt
```diff
@@ -139,6 +141,18 @@ open class RebarBlock private constructor(val block: Block) : WailaSupplier, Key
...
+        editBlockData(BlockData::class.java, editor, applyPhysics)
+    }
+
+    @JvmOverloads
+    fun <D : BlockData> editBlockData(dataType: Class<D>, editor: Consumer<D>, applyPhysics: Boolean = true) {
+        val blockData = block.blockData
+        editor.accept(dataType.cast(blockData))
+        block.setBlockData(blockData, applyPhysics)
+    }
+
```
LordIdra, line 154: Worth making these PylonUtils so they can be used on normal blocks too?

rebar/src/main/kotlin/io/github/pylonmc/rebar/util/RebarUtils.kt
```diff
@@ -816,4 +816,34 @@ fun Material.getPreferredTool(): Material? {
...
+    item.add(amount)
+    notifyWindows(slot)
+}
+
+fun VirtualInventory.unsafeSubtract(slot: Int, amount: Int) {
+    val item = getUnsafeItem(slot)!!
+    item.subtract(amount)
+    if (item.isEmpty) unsafeItems[slot] = null
+    notifyWindows(slot)
+}
```
LordIdra, line 906: Worth a short doc explaining what `unsafe` means / when to use this over `addItemAmount` etc, if someone comes across this?

rebar/src/main/kotlin/io/github/pylonmc/rebar/util/RebarUtils.kt
```diff
@@ -816,4 +816,34 @@ fun Material.getPreferredTool(): Material? {
...
+
+fun VirtualInventory.unsafeAdd(slot: Int, amount: Int) {
+    val item = getUnsafeItem(slot)!!
+    item.add(amount)
+    notifyWindows(slot)
+}
+
+fun VirtualInventory.unsafeSubtract(slot: Int, amount: Int) {
```
LordIdra, line 900: Would it be worth a check here to make sure that amount is less or equal to the amount already there

JustAHuman-xD: personally I would say no, these methods already should only be used in very few circumstances and only when you know for sure what you're doing, including that the item has at least the amount provided.

LordIdra (approved):

PR #863, opened 2026-06-30
Simple processor implementation

LordIdra:
- Prevents `facing` error in `getBlockTextureProperties` from propagating up the stack
- Adds try/catch to code that obtains constructors to provide more informative error message
- Adds `RebarProcessor`, a replacement for `ProcessorRebarBlock`. This is not finished yet and intended to be integrated gradually as more features are added
- Adds state checks for `ProgressBar#fuelRemaining` for better error messages
- Adds `valueStyle` to `UnitFormat`


rebar/src/main/kotlin/io/github/pylonmc/rebar/processor/RebarProcessor.kt
```diff
@@ -0,0 +1,76 @@
+package io.github.pylonmc.rebar.processor
+
+class RebarProcessor {
```
Seggan, line 9: Should probably mark this as unfinished

Seggan (approved):

balugaq (approved):

PR #870, opened 2026-07-07
More unit API

Seggan:
- Added proper i18n for unit names
- Cleaned up `UnitFormat` a bit
- Improved docs
- Changed the confusing behavior of `autoSelectPrefix` and renamed it
- Finally fixed the percent sign having a space before it


balugaq (approved):

rebar/src/main/kotlin/io/github/pylonmc/rebar/util/gui/unit/MetricPrefix.kt
```diff
@@ -7,7 +7,12 @@ import net.kyori.adventure.text.Component
  *
  * Enum order is guaranteed to be the same as the SI unit prefixes, from largest to smallest.
  */
-enum class MetricPrefix(val scale: Int) {
+enum class MetricPrefix(
+    /**
+     * Scale, where `10^scale` is the multiplier of the unit. For example, for kilo, scale is 3, because 1 kilogram is 10^3 grams.
+     */
+    val scale: Int
```
OhmV-IR, line 14: Perhaps order of magnitude is a better name for this?

rebar/src/main/kotlin/io/github/pylonmc/rebar/util/gui/unit/MetricPrefix.kt
```diff
@@ -35,6 +40,14 @@ enum class MetricPrefix(val scale: Int) {
...
-    val abbreviation: Component = Component.translatable("rebar.unit.prefix.${name.lowercase()}.abbr")
+    val translationKey: Component = Component.translatable("rebar.unit.prefix.${name.lowercase()}.name")
+    val abbreviationKey: Component = Component.translatable("rebar.unit.prefix.${name.lowercase()}.abbr")
+
+    companion object {
+        /**
+         * [HECTO], [DECA], and [DECI]
+         */
+        @JvmField
+        val COMMONLY_UNUSED_PREFIXES = setOf(HECTO, DECA, DECI)
```
OhmV-IR, line 51: ? Quetta is not commonly used either, why specifically these ones in the set. Also I don't see where this is used, if it's not used internally just yeet this. (On phone so I can't check for sure if it's used or not)

rebar/src/main/kotlin/io/github/pylonmc/rebar/util/gui/unit/UnitFormat.kt
```diff
@@ -1,70 +1,141 @@
...
+
+    /**
+     * Returns a **new** [UnitFormat] with the same parameters as this one but with a different default prefix
+     */
+    fun withDefaultPrefix(prefix: MetricPrefix) = UnitFormat(singular, plural, abbreviation, noSpace, prefix, defaultStyle)
+
+    /**
+     * Returns a **new** [UnitFormat] with the same parameters as this one but with a different default style
+     */
+    fun withDefaultStyle(style: Style) = UnitFormat(singular, plural, abbreviation, noSpace, defaultPrefix, style)
```
OhmV-IR, line 54: Instead of adding a function for this with changing a different parameter every time why no clone method?

rebar/src/main/kotlin/io/github/pylonmc/rebar/util/gui/unit/UnitFormat.kt
```diff
@@ -1,70 +1,141 @@
...
 
-    fun format(value: Long) = format(BigDecimal.valueOf(value))
+    fun format(value: Long) = format(value.toBigDecimal())
 
-    fun format(value: Float): Formatted {
-        check(!value.isNaN() && !value.isInfinite()) { "Cannot format NaN or infinite values" }
-        return format(BigDecimal.valueOf(value.toDouble()))
-    }
+    /**
+     * NaN and infinity are not supported
```
OhmV-IR, line 129: Also null

rebar/src/main/kotlin/io/github/pylonmc/rebar/util/gui/unit/UnitFormat.kt
```diff
@@ -1,70 +1,141 @@
...
+     */
+    fun format(value: Float) = format(value.toDouble())
 
+    /**
+     * NaN and infinity are not supported
+     */
     fun format(value: Double): Formatted {
-        check(!value.isNaN() && !value.isInfinite()) { "Cannot format NaN or infinite values" }
-        return format(BigDecimal.valueOf(value))
+        require(!value.isNaN() && !value.isInfinite()) { "Cannot format NaN or infinite values" }
```
OhmV-IR, line 137: You could check for null here to have a better error message but it throws quickly either way so doesn't matter much.

rebar/src/main/kotlin/io/github/pylonmc/rebar/util/gui/unit/UnitFormat.kt
```diff
@@ -83,23 +154,26 @@ class UnitFormat @JvmOverloads constructor(
         private val badPrefixes = EnumSet.noneOf(MetricPrefix::class.java)
 
         /**
-         * Sets the number of significant figures. For example, if this is set to 3, then a value
-         * of 0.472894 will be shown as 0.473.
+         * Sets the number of significant figures. Uses [RoundingMode.HALF_UP] for rounding.
+         * For example, if this is set to `3`, then a value of `0.472894` will be shown as `0.473`.
```
OhmV-IR, line 158: This is a bad example to use as the result would be the same with a 3 DP round. An example where the result is different than a DP round and you showcase the half up rounding is best imo. Ie 3.755 -> 3.76

rebar/src/main/kotlin/io/github/pylonmc/rebar/util/gui/unit/UnitFormat.kt
```diff
@@ -83,23 +154,26 @@ class UnitFormat @JvmOverloads constructor(
...
          */
         fun significantFigures(sigFigs: Int) = apply { this.sigFigs = sigFigs }
 
         /**
          * Sets the number of decimal places. This overrides significant figures if both are set.
+         * If [force] is true, the formatted number will always have this many decimal places.
```
OhmV-IR, line 164: Could add output example here as well but I'm not requiring it.

rebar/src/main/kotlin/io/github/pylonmc/rebar/util/gui/unit/UnitFormat.kt
```diff
@@ -117,235 +191,241 @@ class UnitFormat @JvmOverloads constructor(
         fun valueStyle(style: Style) = apply { this.valueStyle = style }
 
         /**
-         * Overrides the default prefix (and adjusts the value shown accordingly).
+         * Overrides the default prefix. **This will not rescale the number like [selectPrefixAndRescale] does.**
          */
         fun prefix(prefix: MetricPrefix) = apply { this.prefix = prefix }
```
OhmV-IR, line 196: Eeek can we make this private or internal somehow? This is a very easy place for someone to not read and make a funny mistake.

rebar/src/main/kotlin/io/github/pylonmc/rebar/util/gui/unit/UnitFormat.kt
```diff
@@ -117,235 +191,241 @@ class UnitFormat @JvmOverloads constructor(
...
             unit = if (abbreviate && abbreviation != null) {
                 unit
-                    .append(usedPrefix.abbreviation)
+                    .append(if (noSpace) Component.empty() else Component.text(" "))
+                    .append(usedPrefix.abbreviationKey)
                     .append(abbreviation)
             } else {
                 unit
-                    .append(usedPrefix.fullName)
+                    .append(Component.text(" "))
```
OhmV-IR, line 248: Not respecting the no space flag

OhmV-IR (changes requested):

rebar/src/main/kotlin/io/github/pylonmc/rebar/util/gui/unit/MetricPrefix.kt
Seggan: This reflects the usage in BigDecmial, also I've every only heard order of magnitude in a comparative sense, not absolute

rebar/src/main/kotlin/io/github/pylonmc/rebar/util/gui/unit/UnitFormat.kt
Seggan: wdym? These are literally added functions for that

rebar/src/main/kotlin/io/github/pylonmc/rebar/util/gui/unit/UnitFormat.kt
Seggan: Null cannot be passed in

rebar/src/main/kotlin/io/github/pylonmc/rebar/util/gui/unit/UnitFormat.kt
Seggan: Null cannot be passed in

rebar/src/main/kotlin/io/github/pylonmc/rebar/util/gui/unit/UnitFormat.kt
Seggan: Y? I found the previous way it did confusing when I was doing electricity. I mean I can add an overload to rescale as well

rebar/src/main/kotlin/io/github/pylonmc/rebar/util/gui/unit/UnitFormat.kt
Seggan: No space flag only works on the abbreviation, otherwise you have `32percent`

rebar/src/main/kotlin/io/github/pylonmc/rebar/util/gui/unit/MetricPrefix.kt
Seggan: Quetta is commonly unseen because numbers of that scale are comparatively rare compared to 100, 10, and 0.1

Also it was introduced like 3 years ago

rebar/src/main/kotlin/io/github/pylonmc/rebar/util/gui/unit/UnitFormat.kt
OhmV-IR: Whoops forgot kt is non nullable by default

rebar/src/main/kotlin/io/github/pylonmc/rebar/util/gui/unit/MetricPrefix.kt
OhmV-IR: Oh you're using this to try to avoid using hecta when you get 100 blocks? Sure ig it kinda works.

rebar/src/main/kotlin/io/github/pylonmc/rebar/util/gui/unit/UnitFormat.kt
OhmV-IR: No I mean instead of unit.withDefaultPrefix(prefix)

   unit.clone()
   .defaultPrefix(prefix)

But there's nothing super wrong with it, I just prefer it being more clear when a clone is happening

JustAHuman-xD: You’d have to change it from immutable to mutable which opens a can of worms in terms of mutation of constants which I think is a worse option

OhmV-IR: Yeah nvm

rebar/src/main/kotlin/io/github/pylonmc/rebar/util/gui/unit/UnitFormat.kt
OhmV-IR: I guess it does make sense with the following:
UnitFormat.BUCKETS.prefix(KILO).format(123)

Which is the primary use case so nvm its fine

rebar/src/main/kotlin/io/github/pylonmc/rebar/util/gui/unit/MetricPrefix.kt
LordIdra: I think this should be documented since as Ohm says it doesn't really make sense without that context

rebar/src/main/kotlin/io/github/pylonmc/rebar/util/gui/unit/UnitFormat.kt
```diff
@@ -1,70 +1,122 @@
...
-    init {
-        allUnits[name] = this
-    }
+    /**
+     * Enables the use of this unit in the custom `<unit:[name]>` tag in [Rebar's custom MiniMessage parser][io.github.pylonmc.rebar.i18n.customMiniMessage]
+     *
+     * @param name the name to be used in the tag
+     * @return this [UnitFormat]
+     */
+    fun allowUseInUnitTag(name: String) = apply { namedUnits[name] = this }
```
LordIdra: Wouldn't it make more sense to it be used by default and provide an option to disable it? Otherwise I can see it being confusing why your new unit doesn't work with the tag

rebar/src/main/kotlin/io/github/pylonmc/rebar/util/gui/unit/UnitFormat.kt
```diff
@@ -1,70 +1,122 @@
...
+    fun format(value: Float) = format(value.toDouble())
 
+    /**
+     * NaN and infinity are not supported
+     */
     fun format(value: Double): Formatted {
-        check(!value.isNaN() && !value.isInfinite()) { "Cannot format NaN or infinite values" }
-        return format(BigDecimal.valueOf(value))
+        require(!value.isNaN() && !value.isInfinite()) { "Cannot format NaN or infinite values" }
+        return format(value.toBigDecimal())
```
LordIdra: While we're at it, would it be worth instead just displaying `NaN` or `Infinity` (maybe in red text) without a unit? Given how often `format` can be called, if something ever goes wrong the check will spam server consoles, and just displaying `NaN` or `Infinity` should make the issue as obvious (if not more obvious)

rebar/src/main/kotlin/io/github/pylonmc/rebar/util/gui/unit/UnitFormat.kt
```diff
@@ -83,23 +135,26 @@ class UnitFormat @JvmOverloads constructor(
...
         /**
-         * Sets whether a value should always have decimal places. For example, if set to true, then
-         * the value 145 will be displayed as '145.0'
-         *
-         * This overrides decimal places if both are set.
+         * Sets the number of decimal places. This overrides significant figures if both are set.
+         * Uses [RoundingMode.HALF_UP] for rounding.
          */
-        fun forceDecimalPlaces(force: Boolean) = apply { this.forceDecimalPlaces = force }
+        fun decimalPlaces(decimalPlaces: Int) = decimalPlaces(decimalPlaces, false)
```
LordIdra, line 202: Nit but can be an overload

rebar/src/main/kotlin/io/github/pylonmc/rebar/util/gui/unit/UnitFormat.kt
```diff
@@ -117,235 +172,310 @@ class UnitFormat @JvmOverloads constructor(
...
+         * [selectPrefixAndRescale] will not use any of these [prefixes] when automatically selecting a prefix.
          */
         fun ignorePrefixes(vararg prefixes: MetricPrefix) = apply { badPrefixes.addAll(prefixes) }
 
         /**
-         * Sets whether the prefix should be automatically selected instead of using the default
-         * prefix (if set).
+         * Same as [ignorePrefixes] but for [MetricPrefix.COMMONLY_UNUSED_PREFIXES]
+         */
+        fun ignoreCommonlyUnusedPrefixes() = ignorePrefixes(MetricPrefix.COMMONLY_UNUSED_PREFIXES)
```
LordIdra: How come this isn't the default? I would have thought we'd want to ignore eg hectoblocks by default.

rebar/src/main/kotlin/io/github/pylonmc/rebar/util/gui/unit/UnitFormat.kt
```diff
@@ -117,235 +172,310 @@ class UnitFormat @JvmOverloads constructor(
...
+                Rebar.translator.canTranslate(abbrKey, Rebar.defaultLanguage)
+            }
+            return UnitFormat(
+                forms = PluralForm.entries.associateWith { Component.text("rebar.unit.$name.$it") },
+                abbreviation = abbr,
+                abbrFormat = abbrFormat,
+                defaultPrefix = prefix,
+                defaultStyle = style,
+            ).allowUseInUnitTag(name)
+        }
```
LordIdra: I feel like the location of units should be standardised across addons (by default) rather than each addon having to specify the location of name/abbreviations. Would it be worth removing/adapting this method and passing an addon to the UnitFormat constructor so the name/abbreviation locations can be automatically determined?

rebar/src/main/kotlin/io/github/pylonmc/rebar/util/gui/unit/UnitFormat.kt
```diff
@@ -1,70 +1,122 @@
...
+     * `v` is replaced with the value (ex 3 to make `"3 u"`), and `u` is replaced with the unit (ex `"3 seggans"`)
+     * @param abbrFormat Same as [format] but `u` is replaced with the abbreviated unit instead. Separate from [format] in order to allow things
+     * like dollars (`$3`, using format string `"uv"`) or percent (`3%`, using format string `"vu"`). Default is `"v u"``.
+     * @param defaultPrefix The prefix (kilo, nano, etc) used for this unit unless specified while formatting.
+     * For example, if you create a 'grams' unit and specify [MetricPrefix.KILO] as the default prefix, calling
+     * [format] with 100 will return '100 kilograms'
+     * @param defaultStyle The style to apply to the unit (not the value) to the output.
+     */
+    @JvmOverloads
+    constructor(
```
LordIdra, line 66: This constructor has a lot of parameters now and looking at the rebar unit definitions, often we want to pick out a specific one. This is fine in kotlin since we can do e.g. `defaultPrefix = [...]`, but in Java this could be a real issue. Can we add builder-style methods which allow you to call e.g. `x = UnitFormat(base, abbreviation).defaultPrefix(...)`?

rebar/build.gradle.kts
```diff
@@ -46,6 +47,7 @@ dependencies {
     implementation("info.debatty:java-string-similarity:2.0.0")
     implementation("org.bstats:bstats-bukkit:2.2.1")
     paperLibrary("com.github.ben-manes.caffeine:caffeine:3.2.2")
+    paperLibrary("com.ibm.icu:icu4j:78.3")
```
LordIdra, line 50: I don't think this was used anywhere in the end?

LordIdra (changes requested):
Changes broadly look good and nice work on the documentation

rebar/src/main/kotlin/io/github/pylonmc/rebar/util/gui/unit/UnitFormat.kt
Seggan: Left it not for documentation purposes

rebar/src/main/kotlin/io/github/pylonmc/rebar/util/gui/unit/UnitFormat.kt
Seggan: See the secondary constructor for UnitFormat

rebar/src/main/kotlin/io/github/pylonmc/rebar/util/gui/unit/UnitFormat.kt
Seggan: This alr exists

rebar/build.gradle.kts
Seggan: It is, see `UnitFormat.Formatted#build`

rebar/src/main/kotlin/io/github/pylonmc/rebar/util/gui/unit/UnitFormat.kt
Seggan: Cause units don't inteinsically have names anymore

rebar/src/main/kotlin/io/github/pylonmc/rebar/util/gui/unit/UnitFormat.kt
LordIdra: Huh why don't we call that constructor in `rebar` then?

Seggan: Cause I wrote that constructor after I wrote this method lol

Seggan: Also cause this method determines wether the abbreviation exists from the translator

rebar/src/main/kotlin/io/github/pylonmc/rebar/util/gui/unit/UnitFormat.kt
LordIdra: How come?

rebar/src/main/kotlin/io/github/pylonmc/rebar/util/gui/unit/UnitFormat.kt
LordIdra: > Also cause this method determines wether the abbreviation exists from the translator

Is that an issue?

rebar/src/main/kotlin/io/github/pylonmc/rebar/util/gui/unit/UnitFormat.kt
LordIdra: There are two `with` methods but there aren't any for most of the fields

rebar/src/main/kotlin/io/github/pylonmc/rebar/util/gui/unit/UnitFormat.kt
Seggan: Side effect of the derived units thing that I never changed back

rebar/src/main/kotlin/io/github/pylonmc/rebar/util/gui/unit/UnitFormat.kt
Seggan: No not really, just smth I never thought of

rebar/src/main/kotlin/io/github/pylonmc/rebar/util/gui/unit/UnitFormat.kt
LordIdra: Is it worth keeping for this? Idk if it would have other use cases but it seems like the kind of thing that might be useful to have

Seggan: Sure

LordIdra (approved):

PR #883, opened 2026-07-24
Fix fluid pipe interactions with pedestal and silo

### LINKED ISSUE #884: silos wont accept pipes
**Issue reported by: averageunusualuser:**
[View original discord thread](https://discord.com/channels/1329177304857055273/1492865446150537227)
when rightclick or shift right clicking it will just start the pipe placing thing

### LINKED ISSUE #885: pipes can't be placed in the pedestal
**Issue reported by: lwj_666:**
[View original discord thread](https://discord.com/channels/1329177304857055273/1510213300212994158)
.

balugaq:


balugaq:
Tested.

balugaq:
The bug happens every time holds the pipe to interact the pedestal, and pipe cancel item use, then pedestal quit the handle.
Similar to silos, there may more blocks like them which pipes cannot interacts with. So I changed pipe's priority instead of theirs. Not sure if there's a side-effect.

QQ_1784897776001

LordIdra (approved):
We really need to come up with a better solution for handling interact priority, but this seems ok for now.

Tested

PR #887, opened 2026-08-09
Update to 26.2

glomdom:


glomdom:
above commit makes it so the versions checked are by part and not by string
so `26.2.0` matches `26.2` by dropping the patch number, and `26.2.1` will be matched with `26.2.1`

Seggan (approved):

PR #890, opened 2026-08-18
Fix KeyedPersistentDataType cross-plugin API

wickidcow:
Adds a Java Function<NamespacedKey, T> overload for KeyedPersistentDataType.keyedTypeFrom.

I encountered a LinkageError when Pylon and Rebar were loaded through separate plugin classloaders. The existing public overload exposes Kotlin's Function1 in the cross-plugin method signature.

The Java Function overload avoids putting a Kotlin runtime type in that API boundary while retaining the existing Kotlin overload for source compatibility.

### DIFF
--- a/rebar/src/main/kotlin/io/github/pylonmc/rebar/datatypes/KeyedPersistentDataType.kt
+++ b/rebar/src/main/kotlin/io/github/pylonmc/rebar/datatypes/KeyedPersistentDataType.kt
@@ -4,6 +4,7 @@ import org.bukkit.Keyed
 import org.bukkit.NamespacedKey
 import org.bukkit.persistence.PersistentDataAdapterContext
 import org.bukkit.persistence.PersistentDataType
+import java.util.function.Function
 
 /**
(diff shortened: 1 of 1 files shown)


rebar/src/main/kotlin/io/github/pylonmc/rebar/datatypes/KeyedPersistentDataType.kt
```diff
@@ -35,7 +36,15 @@ abstract class KeyedPersistentDataType<T : Keyed>(val type: Class<T>) : Persiste
...
+        /**
+         * Kotlin-source compatibility overload.
+         *
+         * This method intentionally remains in the JVM ABI for already compiled Kotlin addons, but
+         * is hidden from Java source so Java plugins do not put kotlin.jvm.functions.Function1 in
+         * their cross-plugin call signature.
+         */
         @JvmStatic
+        @JvmSynthetic
         fun <T : Keyed> keyedTypeFrom(
```
Seggan, line 48: It's ok Rebar is unstable we can break API/ABI, u can remove the kt method

PR #893, opened 2026-08-18
Restore recipes after resource reload

wickidcow:
Fixes Rebar/Pylon recipes disappearing after Paper reloads server resources.

I ran into this while testing crafting on a live Paper 26.2 server. Paper rebuilds its recipe registry during a resource/datapack reload, which can remove recipes that Rebar registered dynamically after startup.

This listens for ServerResourcesReloadedEvent and re-registers Rebar recipes that are missing from the live registry.

Vanilla recipes that Rebar discovered during startup are left alone.

### DIFF
--- a/rebar/src/main/kotlin/io/github/pylonmc/rebar/recipe/RecipeType.kt
+++ b/rebar/src/main/kotlin/io/github/pylonmc/rebar/recipe/RecipeType.kt
@@ -1,5 +1,6 @@
 package io.github.pylonmc.rebar.recipe
 
+import io.github.pylonmc.rebar.nms.NmsAccessor
 import io.github.pylonmc.rebar.recipe.vanilla.*
 import io.github.pylonmc.rebar.registry.RebarRegistry
 import org.bukkit.Bukkit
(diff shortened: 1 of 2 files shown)


rebar/src/main/kotlin/io/github/pylonmc/rebar/recipe/RecipeType.kt
```diff
@@ -153,6 +154,57 @@ open class RecipeType<T : RebarRecipe>(private val key: NamespacedKey) : Keyed,
...
+            queueDynamic(VANILLA_SMELTING.recipes)
+            queueDynamic(VANILLA_SHAPED.recipes)
+            queueDynamic(VANILLA_SHAPELESS.recipes)
+            queueDynamic(VANILLA_SMITHING_TRANSFORM.recipes)
+            queueDynamic(VANILLA_SMITHING_TRIM.recipes)
+            queueDynamic(VANILLA_SMOKING.recipes)
+
+            for (recipe in DUMMY_CRAFTING.recipes) queueIfMissing(recipe)
+            for (recipe in DUMMY_COOKING.recipes) queueIfMissing(recipe)
+            for (recipe in DUMMY_SMITHING.recipes) queueIfMissing(recipe)
```
Seggan, line 199: Braces plz

Seggan (commented):
Untested

PR #896, opened 2026-08-19
fix(i18n): fall back to default language for missing keys

### LINKED ISSUE #591: Allow patched translations
For example:

`en.yml`
```yml
foo:
	bar: "Baz"
	eggs: "Spam"
```

`en_BR.yml`
```yml
foo:
	bar: "Bauz"
```

In `en_BR`, `foo.bar` is `Bauz`, but in `en_US` it is `Baz`. Both languages return `Spam` for `foo.eggs`.

ybw0014:
Closes #591

Missing a translation key no longer shows the full key, it fallbacks to the default language instead. 
The console warning for the missing key is kept.

Existing entry:
Image_2026-08-19_00-36-50_hzfbivnq 5xt

New entry added, without Chinese translation (current behavior):
image

New behavior, fallback to English:
Image_2026-08-19_00-38-06_qcuj2k4p cmx

### DIFF
--- a/rebar/src/main/kotlin/io/github/pylonmc/rebar/addon/RebarAddon.kt
+++ b/rebar/src/main/kotlin/io/github/pylonmc/rebar/addon/RebarAddon.kt
@@ -82,7 +82,7 @@ interface RebarAddon : Keyed {
         RebarRegistry.ADDONS.register(this)
         if (!suppressAddonNameWarning) {
             for (locale in translator.languages) {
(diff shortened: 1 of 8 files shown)


Seggan (approved):
I'd eventually want a more comprehensive fallback system (ex `en-BR` falling back to `en`) but this works for now

LordIdra (approved):
Thanks for the PR

PR #906, opened 2026-09-02
Make clazz the second parameter of BlockStorage.getAs

### LINKED ISSUE #904: Swap parameter order of getAs in BlockStorage
This is a bit of a nitpick but nice to keep consistent - as convention we usually have the clazz parameter second - e.g. `fun isRebarItem(stack: ItemStack?, clazz: Class<*>): Boolean` as opposed to `fun <T> getAs(clazz: Class<T>, blockPosition: BlockPosition?): T?`. Would be nice to swap the parameter order to match this in the BlockStorage getAs methods

chrisriv10:
Fixes #904

Updates the BlockStorage.getAs method signatures so the clazz parameter is the second argument for consistency with the rest of the codebase. Updates affected Kotlin and Java call sites while preserving existing behavior.

Validation:

* :rebar:compileKotlin
* :test:compileJava
* :rebar:classes :test:classes

All completed successfully.

### DIFF
--- a/rebar/src/main/kotlin/io/github/pylonmc/rebar/block/BlockStorage.kt
+++ b/rebar/src/main/kotlin/io/github/pylonmc/rebar/block/BlockStorage.kt
@@ -129,7 +129,7 @@ object BlockStorage : Listener {
      * @throws IllegalArgumentException if the chunk containing the block is not loaded
      */
     @JvmStatic
-    fun <T> getAs(clazz: Class<T>, blockPosition: BlockPosition?): T? {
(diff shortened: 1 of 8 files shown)


LordIdra (changes requested):
Since we are still in development and so making breaking changes constantly, it is fine to remove the methods rather than deprecate them so that we don't have old methods cluttering up the codebase.

When you remove the old methods, you should also check if Pylon still compiles (use ./gradlew runServer in parallel-dev-repo) as there might be code in there which uses the old methods. In which case you should also update those methods and submit a corresponding Pylon pull request.

PR #907, opened 2026-09-02
Error when commands receive nonexistent player names

### LINKED ISSUE #788: Commands do not error when given nonexistent player names
/rb give does this
Image

/rb research add Idraq pylon:assembling just does not give any feedback at all

Have not tested others but all should be checked probably

chrisriv10:
Fixes #788

Updates commands using multi-player arguments so targets that resolve to no players produce a command error instead of silently succeeding. Existing behavior for valid player targets is unchanged.

Validation:

* :rebar:compileKotlin
* :rebar:classes :test:classes

Both completed successfully.


LordIdra (changes requested):
I believe that this will throw errors if you run the command with a selector and no entities are in range, which would display the incorrect error message.

I would assume there is a better way to do this as returning an empty list for incorrect player names just seems like strange/incorrect behaviour on Brigadier's part? But I'm not really sure, it might be worth investigating. Either way this should be modified so that it only errors on genuine misspellings rather than just empty selector results

PR #908, opened 2026-09-02
Default UnitFormat floating-point values to 2 decimal places

### LINKED ISSUE #694: make Celsius UnitFormat go to 1 decimal place by default
there's no reason it shouldn't be default, currently it just displays like 10 dp lol

chrisriv10:
Fixes #694

Updates UnitFormat so Float and Double values default to at most two decimal places instead of exposing excessive floating-point precision.

Trailing zeros continue to be stripped, integer and BigDecimal formatting remain unchanged, and explicit decimalPlaces/significantFigures settings continue to take precedence.

Validation:
- :rebar:compileKotlin
- :rebar:test --tests "*UnitFormat*"

Both completed successfully.

### DIFF
--- a/rebar/src/main/kotlin/io/github/pylonmc/rebar/util/gui/unit/UnitFormat.kt
+++ b/rebar/src/main/kotlin/io/github/pylonmc/rebar/util/gui/unit/UnitFormat.kt
@@ -146,6 +146,10 @@ class UnitFormat @JvmOverloads constructor(
 
     fun format(value: Float) = format(value.toDouble())
 
+    /**
+     * Formats a [Double] value. Unless rounding is configured with [Formatted.decimalPlaces] or
+     * [Formatted.significantFigures], the value defaults to at most 2 decimal places.
+     */
     fun format(value: Double): Formatted {
         return Formatted(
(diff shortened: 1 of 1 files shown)


LordIdra (changes requested):
Thanks for the PR!

I don't think that we need the three new fields. Literally just defaulting `decimalPlaces` to 2dp instead of calling `BigDecimal#scale` should be fine as that is a more sensible default. @Seggan any thoughts?

We do not use unit tests for this project for various architectural and tooling reasons. Running `./gradlew test` in the rebar directory does not seem to output any test results. How are you running the tests?

Please do have a read of the contributing docs (https://pylonmc.github.io/documentation/contributing/getting-started/) if you have not already as they contain a lot of useful info

Finally, can I ask if you used AI for this pull request?

Seggan (changes requested):
> I don't think that we need the three new fields. Literally just defaulting decimalPlaces to 2dp instead of calling BigDecimal#scale should be fine as that is a more sensible default. @Seggan any thoughts?

Yes correct.

Also this is now broken with the most recent changes.

